PR #5, opened 2018-11-14
Added Princeton V3 met data reader

dpsarmie:
Added code to the current Princeton reader in LIS that allows for the reader to ingest Princeton Version 3 met driver data.

The new reader now requires the user to define the version of the Princeton dataset that they are using. The old reader's capabilities of reading V2 and V2.2 Princeton data are not changed with the addition of the V3 reader.

Test case configuration file is located at /discover/nobackup/dpsarmie/config_RefET_Princeton.cfg

(Let me know if anything went wrong during the git pushing/pulling process!)

### DIFF
--- a/lis/metforcing/princeton/0Intro_princeton.txt
+++ b/lis/metforcing/princeton/0Intro_princeton.txt
@@ -9,7 +9,7 @@
 !BOP
 !\section{Princeton}
 !This is the 50 year forcing data (Sheffield et al., 2006) produced at Princeton 
-!University. 
+!University.
 !Versions 2 and 2.2 of the data are global at 1 degree spatial resolution, at 3 hourly 
 !intervals. 
 !Version 3 is at 0.25 degrees and is psedo-global (southern boundary at 59.875 deg S), at 
--- a/lis/metforcing/princeton/0Intro_princeton.txt
+++ b/lis/metforcing/princeton/0Intro_princeton.txt
@@ -8,7 +8,11 @@
 !
 !BOP
 !\section{Princeton}
-!This is the 50 year forcing data~\cite{princeton} produced at Princeton 
-!University. The data is global at 1 degree spatial resolution, at 3 hourly 
+!This is the 50 year forcing data (Sheffield et al., 2006) produced at Princeton 
+!University. 
+!Versions 2 and 2.2 of the data are global at 1 degree spatial resolution, at 3 hourly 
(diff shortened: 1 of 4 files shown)


jvgeiger:
Would you please point me to some documentation for the Princeton v3 datasets?  I would like to learn more about why this new version has missing data.

The Princeton v2 data are complete datasets, meaning they provide all the required input met forcing fields to drive a land surface model.  The reader aborts if there are missing data because the LSM in LIS expect all land grid-cells to have forcing data.

I see that your config_RefET_Princeton.cfg is running only the template water model.  If you had enabled an LSM like Noah 3.6, it would have crashed because of the missing forcing.

dpsarmie:
We have been trying to locate some sort of documentation for version 3 without much success. Kristi was going to email some folks at Princeton to see if they had it available. I'll check in with her to see if they got back to her.

The catch for the 'missing nearest neighbor' seemed important, which is why I didn't delete it altogether. I can revert that specific change if you'd like. My runs started from 2000 and went up to ~Oct 2014 without any issues. It wasn't until late 2014 that I encountered the missing data problem and changed the code for the error catch, so I'm hoping that most of the available dataset is unaffected by that error.

Let me know how you want to proceed. I'll keep trying to get some sort of documentation and I'll let you know if I find it. Thanks!

jvgeiger:
Has anyone contacted Princeton about the files with missing data?  Perhaps Princeton needs to reprocess those files.

jvgeiger:
Would you please change line 162 to use princeton_struc(n), simply to be consistent?

Also would you update the 0Intro_princeton.txt file to describe v2, v2.2, and v.3, including pointing out that v3 does not include precipitation?

jvgeiger:
For now, let's rework the "missing nearest neighbor" check, so that versions 2 and 2.2 will stop and version 3 will assign LIS_rc%udef.

However, instead of an actual stop statement, please "call LIS_endrun".

Thanks.

dpsarmie:
I've gone ahead and added the changes that were suggested. Let me know if there's anything else that should get changed.

dpsarmie:
Hi Jim,

I went ahead and created a test case using the documentation you showed me. All the files are in the following directory: /discover/nobackup/dpsarmie/work/gitLIS/PrincetonReader/princetonV3

The directions were pretty straightforward and easy to follow! Let me know how it goes when you get around to it.

PR #31, opened 2018-12-14
IMERG LIS reader filename update

dpsarmie:
This code modification updates the LIS reader to read in version
V05B (latest IMERG algorithm version available) instead of V03D.
This code has also added a config option to allow the user to switch
between the three release versions of IMERG (early, late, and final),
all of which have different filename conventions.
 If no version is declared in the configuration file, the reader defaults to the final
version.

### DIFF
--- a/lis/configs/lis.config.adoc
+++ b/lis/configs/lis.config.adoc
@@ -5812,9 +5812,15 @@ anchor:sssec_supp_imerg[IMERG precipitation]
 `IMERG forcing directory:` specifies the location of the
 GPM IMERG precipitation forcing files.
 
+'IMERG version:' specifies the version of the GPM IMERG 
+precipitation forcing files. The valid options are: early,
+late, and final. If no version is specified, the reader will
+default to the 'final' version of GPM IMERG. 
+
 .Example _lis.config_ entry
 ....
-MERG forcing directory:     ./FORCING/IMERG
+IMERG forcing directory:     ./FORCING/IMERG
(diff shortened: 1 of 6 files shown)


jvgeiger:
Thanks for the update.  Would please make a formatting update?

In lis.config.adoc, would you replace 'IMERG version:' with `IMERG version:`.  In case that is hard to read, I replaced the single quotes with back ticks.  And please quote 'early', 'late', and 'final'.

jvgeiger:
I see that you changed the file names from V03D to V05B.  I searched our project disk and could not find any V03D files, but I want to confirm that with you.

Also, is that product/version number something that we should expect to change in the future?  Should we organize the IMERG data on project disk according to that version number like
MET_FORCING/IMERG/V05B/YYYYMM?

jvgeiger:
Would you package a testcase like you did for PRINCETON V3?

dpsarmie:
I'll upload the config changes and the test case early next week.

Yes, the version change is for the latest IMERG release (and matches what is on Discover).  There is an old folder on discover with the older version. I'll reach out to the IMERG team and see how often they update versions to see if the organization method you outlined and adding a config option to the reader are going to be needed.

Thanks for taking a look at the code.

dpsarmie:
I've uploaded a testcase for the IMERG reader here: /discover/nobackup/dpsarmie/work/gitLIS/testcase/imerg

Let me know if there are any issues.

PR #32, opened 2018-12-18
IMERG Precipitation Reader for LVT

dpsarmie:
This update adds a reader to LVT that reads in GPM IMERG precipitation
data. This reader only works with the latest IMERG release (V05B) and
includes a configuration option that allows the user to switch between
the early, late, and final versions of the IMERG product.


dpsarmie:
More IMERG stuff:
config file: /discover/nobackup/dpsarmie/work/gitLIS/config_feature-LVT_IMERG.cfg

Paths to data are in the config file (same paths that are used for the IMERG filename commit). Let me know if you have any other questions.

jvgeiger:
Thanks for pull request.  Would please do a little code clean up?  Would you delete the trailing whitespace at the end of several lines in the IMERG_dataMod.F90 file?

Also would you package up the testcase data?

dpsarmie:
I've added a testcase for this reader here: /discover/nobackup/dpsarmie/work/gitLIS/testcase/imerglvt

Since this is an LVT testcase, I followed most of the instructions posted on the testcase creation guide with some modified steps to conform to LVT. 
Some of the files that aren't used by LVT (forcing_variables.txt, MODEL_OUTPUT_LIST.TBL) are still included but can probably be deleted. I'll leave the final decision to you.

Let me know if you have any issues or if you want the LVT testcase modeled in a different way.

jvgeiger:
I am unable to reproduce your results.  My LVT log is identical to yours, but my results differ.  My run is not processing GDAS the same way.  Your results have a countgdas of 8, mine is 4.  Also my results have 0 rainfgdas.  Would you please rerun /discover/nobackup/dpsarmie/work/gitLIS/testcase/imerglvt

One thing that I noticed was your ctl files and your TARGET_OUTPUT are binary (gs4r), but OUTPUT is written as NetCDF.

fenxi_rainfgdas 1 lvt

dpsarmie:
Hi Jim,
I ran the testcase in /discover/nobackup/dpsarmie/work/gitLIS/testcase/imerglvt and everything looked like it was working correctly (countGDAS of 8). The only thing I changed was the netcdf output to binary in the lvt.config file (now changed in the testcase).

I used LVT compiled from the submitted branch, I also pulled a fresh copy from the main LISF repository and ran it there.  All of these runs were on Intel14 compilers. I can try to run it using other configurations to try and replicate the error if that'll help.

jvgeiger:
Thanks Daniel.  Your test case runs for me under Intel 14, but it produces bad results under Intel 18.  Would you please recompile and rerun using the Intel 18 compilers to see whether this happens for you?

dpsarmie:
I tested the case using the Intel 18 compilers and ran into the same errors you saw. There's actually an issue with the GDAS reader in LVT when it is being compiled using Intel 18.
I did a quick fix to the reader and it is now properly working with the Intel 18 compilers. I'll talk to you offline about the fix and whether or not we should include it.

jvgeiger:
Thanks!

Issue #66, opened 2019-03-04
LDT downscaling for GDAS not working correctly


dpsarmie:
When using the 'Metforce temporal downscaling' option in LDT with GDAS, the GDAS reader does not properly reset and results in the same field until the secondary met forcing source resets.
This was found when trying to downscale CHIRPS data using the GDAS forcing.

dmocko:
Is this issue now closed, with pull request #67 now closed?

dpsarmie:
Yes, this issue is resolved.  I thought that referencing the issue in the commit message would auto-close the linked issue once the commit was merged. Maybe I did something wrong? (Or that isn't a feature on Github.)

dmocko:
Looks like you did things correctly, and I thought the linked issue would auto-close.  Not sure what happened.  Thanks for confirming.

PR #68, opened 2019-03-04
Bug fix for LDT downscaling bug when using GDAS

dpsarmie:
This fix is for the bug that occurred in LDT when using the 'Metforce temporal downscaling'
option along with the GDAS forcing dataset. The GDAS reader was not properly reseting and ingesting
new GDAS data at the downscaled time interval. A secondary flag was added to the reader that corrects this issue. This fix is similar to the fix in the MERRA2 and CHIRPS readers in LDT.

This fix resolves issue #66

### DIFF
--- a/ldt/metforcing/gdas/create_gdasfilename.F90
+++ b/ldt/metforcing/gdas/create_gdasfilename.F90
@@ -182,9 +182,9 @@ subroutine create_gdasfilename(option, name00, name03, name06, &
   F06flag = .false. 
 
   if ( option == 2 ) then !bookend 2
-     if ( .not. is_analysis_hr ) F06flag = .true. ! need to read two files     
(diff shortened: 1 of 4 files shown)


jvgeiger:
Sorry @dpsarmie and @karsenau, I am confused by the update to create_gdasfilename.  The documentation suggests that, when reading bookend 1 at an analysis hour, you also need the F06 file.  So line 187 should be

```
if ( is_analysis_hr ) F06flag = .true. ! need to read two files
```

This is consistent with lis/metforcing/gdas/create_gdasfilename.F90.

Would you please review that one more time?

dpsarmie:
Hey Jim,
You're right, line 187 should be setting the flag as .true.
If you need me to make the modification, let me know.

PR #72, opened 2019-03-13
Add Multi-Radar Multi-Sensor (MRMS) forcing reader

jerlingi:
This update adds the capability to read MRMS forcing as well as an option within the reader that allows the user to input dynamically varying masks to delineate where the radar data should/should not be used.

Test case data can be found at /discover/nobackup/projects/wldas/share/mrms_testcase

This resolves trac tickets #538, #597, #625


jvgeiger:
Would you please give me read permission to your mrms_testcase files?

jerlingi:
Sorry about that! You should have permission now.

jvgeiger:
Hello,

My run of the mrms test case crashes at line 70 in finalize_mrms_grib.F90.

From what I see, the run will use nearest neighbour interpolation for the MRMS data, but LIS is crashing trying to deallocate the array needed for upscaling.

Would you please review the logic in finalize_mrms_grib.F90?

I suspect the problem is in (res.eq.0.01).  Even though res is 0.01, it is not 0.01 w.r.t. floating point.

A simpler approach maybe to just check each array like this:

```
if ( allocated(mrms_grib_struc(n)%rlat1) ) then
   deallocate(mrms_grib_struc(n)%rlat1)
endif
```

PR #127, opened 2019-05-28
Add backward compatibility with ESMF 5.2.0rp3

emkemp:
Changed an INTENT argument setting in write_LDTSIobs.F90 to allow backward compatibility with ESMF 5.2.0rp3.


emkemp:
This pull request was rejected since it targeted merging directly to master.  It was meant for the support/lisf-557ww-7.3 branch.

PR #129, opened 2019-05-29
UA SWE/SNWD Reader for LVT

rkim3:
This update adds a reader to LVT that reads in University of Arizona SWE/ Snow depth data.

I've added test cases for this reader : /discover/nobackup/rkim3/gitLIS/testcase/UA_SWE
                                                                /discover/nobackup/rkim3/gitLIS/testcase/UA_SNWD

Let me know if you have any other questions. Thanks!


dmocko:
Hi,

Thanks much for submitting this pull request for this needed feature.
Thanks also for creating and sharing testcases.

I've started to review and test, and I have three requests of you:

1) My code compiled, but when I ran, I received this error:

****************Error****************************
observation init routine for source UA SWE is not defined
please see the configs/lvt.config.master file.....
program will seg fault.....
****************Error****************************

You likely edited these two files:
    lvt/plugins/LVT_datastream_pluginMod.F90
    lvt/plugins/LVT_pluginIndices.F90

However, they were not in the commit and pull request.
Can you please also commit your edits to these files?

2) Can you please also add the required documentation
   for using "UA SWE" and "UA SNWD" to the this file?
      lvt/configs/lvt.config.master

3) Can you modify both of your testcases to generate
   tar files as described on this page?
      https://github.com/NASA-LIS/LISF/blob/master/docs/howto_create_lis_testcases/howto_create_lis_testcases.adoc

   The important thing is to include all the necessary
   files (but not any more than necessary) into the tar
   files as described.  Don't include any testcase files
   in your commit, as detailed in the link.

Thank you!

David

rkim3:
Hi,

Thanks for your reviewing.
Here are my responses to your requests:

1. I've committed all my edits based on your requests #1 and #2.
   
    lvt/plugins/LVT_datastream_pluginMod.F90
    lvt/plugins/LVT_pluginIndices.F90
    lvt/configs/lvt.config.master

2. I've modified both my testcases with tar files.

Let me know if you have further questions.

Thanks,
Rhae Sung

dmocko:
Hi,

Thanks much for the code commits.  I can now successfully compile LVT.

I didn't realize this until I started running your testcases, but there is only one data file from UA for both the SWE and SnowDepth variables.  Thus, if would be better if there was only one reader for them both, instead of separate instances of "UA_SWE" and "UA_SNWD".

Instead of separate directories in lvt/datastreams, and separate plugins, can you please change to code so there is only "UA_Snow" or a similar name?

Look at other datastreams (such as FLUXNETmte) for how to have LVT process more than variables in a single datastream through LVT.

It would be best if you worked off an updated master branch to add this "feature/UA_Snow" code, as there are now conflicts as well.

Sorry for this trouble, but it will be worth it for users to be able to analyze both SWE and SnowDepth in one LVT instances.

No need for updating any testcase, though.  Please just update and commit the code, and I will take another look at it.

David

dmocko:
Just a short update to say that I'm still reviewing this pull request.  I will have some code commits as well to refine some items.

rkim3:
Thanks, David!

dmocko:
Hi Rhae Sung,

I just committed some updates.  Can you update your code with these updates, and then generate a small testcase?  Just a month or so, not the entire record.  Perhaps across a water year transition.  Please also process both SWE and SnowDepth in the same testcase:

#                  Datastream 1               |                    Datastream 2
# Short   Selected               Time   Total |   Short   Selected               Time   Total
#  Name     Levels  Units    Dir  Avg  Levels |    Name     Levels  Units    Dir  Avg  Levels
LVT datastream attributes table::
SWE            1 1  kg/m2    -     1     1      SWE            1 1  kg/m2    -     1     1
SnowDepth      1 1  m        -     1     1      SnowDepth      1 1  m        -     1     1
::

Bundle them as before, with only including the data necessary to run and evaluate this particular testcase.  After you have them, I will also test, and then we can approve this pull request.

David

rkim3:
Hi David,

It looks like LVT is trying to read LIS output for every hour. This happened even after choosing different computation frequencies (e.g., 3hr, 1da):

 [INFO] LVT cycle time: 09/01/2009 11:00:00
 [WARN] LIS file
 /discover/nobackup/projects/SEUP/rkim3/SEUP/NoahMP/MERRA2/OL/OUTPUT/SURFACEMODE
 L/200909/LIS_HIST_200909011100.d01.nc does not exist
 [INFO] LVT cycle time: 09/01/2009 12:00:00
 [INFO] Reading LIS output
 /discover/nobackup/projects/SEUP/rkim3/SEUP/NoahMP/MERRA2/OL/OUTPUT/SURFACEMODE
 L/200909/LIS_HIST_200909011200.d01.nc
 [INFO] Reading UA data
 /discover/nobackup/projects/lis/STN_DATA/UA_SNOW/4km_SWE_Depth_WY2009_v01.nc
 [INFO] LVT cycle time: 09/01/2009 13:00:00
 [WARN] LIS file
 /discover/nobackup/projects/SEUP/rkim3/SEUP/NoahMP/MERRA2/OL/OUTPUT/SURFACEMODE
 L/200909/LIS_HIST_200909011300.d01.nc does not exist

Could you double check/test this?

Rhae Sung

dmocko:
Hi,

Thanks for reporting this, but I don't think that this is an issue.
LVT often does this when the interval is less than the interval of the LIS output data.

LVT only calculates the comparison metrics for when there is data at the same time. 

What values do you have for "Metrics computation frequency": "Metrics output frequency:", and "LIS output interval:"?

It looks like you only have LIS data on 12Z.

Please check your output, but I suspect that it will be correct as you have expected, based on your settings, despite the "does not exist" messages.

David

rkim3:
Hi,

This is not that case. I have 3hr-interval LIS output and my settings are: 

Metrics computation frequency: 1da
Metrics output frequency: 1da
LIS output interval: 3hr

I also tried to use a '3hr' metrics computation frequency value, but LVT still attempted to read LIS output hourly.

I agree that "this is not an issue because LVT only calculates the comparison metrics for when there is data at the same time". 

Rhae Sung

dmocko:
Rhae Sung,

Thanks for the info.  Based on your settings, I expect that LVT will do the following:

1) Calculate the daily average of the LIS output (the average of the 3-hourly data for that day.
2) Calculate the daily average of the UA SNOW data (which is only available once a day at 12Z).
3) Calculate the METRICS between these daily averages.

If you turned on TS file output in the METRICS file, you should have daily output files of these calculations.

Please compare (outside of LVT) the source datasets against the LVT output to confirm that this is correct.  Or, bundle up the testcase, and I can also check.

If you only want to compare your 12Z LIS output with the 12Z UA SNOW data, instead of daily averages, then you'll need to adjust your settings. We can take that discussion offline.

David

rkim3:
Eric,

Thanks for your explanation. 

I compared the source datasets against the LVT outputs, and those look good. You can also check the testcase here:

/discover/nobackup/rkim3/gitLIS/testcase/test

Rhae Sung

dmocko:
Great!  Thanks, this testcase works for me, and I match your target output.

Since I wrote a portion of the code, I think it's worth having @emkemp also quickly verify this testcase.  If it looks good to him, one of us will approve the pull request.

Issue #136, opened 2019-06-05
LVT cannot compile with gfortran/gcc strict checks


emkemp:
The LVT Config.pl always uses ifort/icc debug flags for strict checks, instead of using flags specific to the selected compiler.  As a result, gfortran/gcc will not compile the program when strict checks are selected.

PR #138, opened 2019-06-05
Added strict check settings for GNU compilers.

emkemp:
Overall logic and flags borrowed from LIS Config.pl.

Fixes issue #136.


emkemp:
Resolves:  #136

Issue #145, opened 2019-06-07
IMERG version number config option


dpsarmie:
The current version of the IMERG reader in LIS is hard-coded to use version V05B. An option should be added to allow the user to switch between different versions of IMERG from the config file.

PR #146, opened 2019-06-07
Add config option to switch IMERG versions

dpsarmie:
This commit will add the a new config option "IMERG product:". The
  previously added "IMERG version:" will now refer to the version
  number (ex. V06B) and the "IMERG product" option will refer to the
  early, late, or final product of IMERG.

This removes the previously hardcoded version of V05B from the IMERG reader.

If these options are not in the config file, the reader will default to V06B
  and 'final' as the IMERG version and product.

A new testcase has also been created for the reader: /discover/nobackup/dpsarmie/work/gitLIS/testcase/imerg_lis/

Resolves: #145


emkemp:
I reviewed the code and tested with both Intel and GNU.  The only bug I see is a typo in the lislog file:

IMERG prodcut : late

should read

IMERG product : late

Can you correct the write statement to fix the typo?

dpsarmie:
Typo in readcrd_imerg.F90 has been fixed

emkemp:
I believe this is ready to merge.  However, since this includes a new test case, the actual acceptance needs to be done by Jim who will add the test case to his collection.

Apologies for the delay. :-(

Issue #8, opened 2018-11-18
Add irrigation interfaces in Noah-MP


sujayvkumar:

jerlingi:
Including updates outlined in W. Nie et al. (2018) (https://agupubs.onlinelibrary.wiley.com/doi/pdf/10.1029/2017WR022178) with the ability to toggle the groundwater abstraction on and off.

dmocko:
Can this issue now be closed?

dmocko:
Closing this issue.  Interfaces were added via pull request #150

PR #150, opened 2019-06-11
Feature/gw irr

jerlingi:
Adds irrigation interfaces to Noah-MP 3.6. This update includes options for specifying the growing season and an option for whether to draw irrigation water from groundwater as in W. Nie et al. 2018 (https://agupubs.onlinelibrary.wiley.com/doi/pdf/10.1029/2017WR022178).

Disclaimer: This update will only work for the default Noah/Noah-MP layering scheme (layering depths of 10 cm, 30 cm, 60 cm, and 100 cm). Future updates will allow for flexibility in soil layering. 

The testcase can be found in /discover/nobackup/projects/wldas/share/noahmp36_irrigation_testcase/

Note: Adding a year's worth of NLDAS-2 forcing to the testcase directory would cause me to go over the file quota for this project, so a symbolic link to Discover files is included in this case.


jerlingi:
Resolves issue #8

emkemp:
Hello:

I just tested the GWoff case, and I'm seeing major differences in the forcing fields one month into the simulation compared to your target output.  (To save disk space, I only output data monthly.)  

I don't see copies of the lis.config files you used to produce your target output.  Can you please point me to them?  I want to make sure they match the samples given in lis/testcases/irrigation/noahmp3.6_gw.

Thanks.

emkemp:
Same story with the GWon case.  Very different results one month in.

emkemp:
DISREGARD EARLIER MESSAGES.

When I run my test again with daily output instead of monthly, my 1 month results for GWoff are very close to yours.  Since the output variables are all tavg, the behavior I experienced earlier is just due to not saving the full time series for the calculations.

I'm still running the GWoff option for the full test period, then will run GWon.

jerlingi:
In case you need them, the copies of the config files can be found in /discover/nobackup/projects/wldas/share/irr_testing_dir/LIS/ (lis.config.GWon and lis.config.GWoff). Please let me know if you have any other questions, and sorry for the delayed response!

emkemp:
Thanks, and don't worry about the delayed response--I figured you were busy.

Do you remember what version of the compiler you used to run your tests?  My results are almost identical, but there is a little bit of difference after several decimal places.

jerlingi:
I used Intel-18.0.3.222 and the environment set by Jim's lis_7_intel_18_0_3_222 module.

emkemp:
Interesting, I'm using the same module.  Do you remember what optimization level you used?  Also, what netCDF4 compression level?

(I'm not greatly concerned about the differences, but I am curious if we can track the reason down.)

jerlingi:
Actually, the gcc compiler might have been switched from /usr/local/other/SLES11.3/gcc/7.3/bin/gcc to /usr/local/other/SLES11.3/gcc/5.3.0/bin/gcc . Let me know if you want me to rerun the target output.

emkemp:
It's okay.  I fiddled around more with my compiler/compression settings, and I'm now getting the same max/min as your target output.  (I'm still running GWon to completion.)

Issue #156, opened 2019-06-19
Missing 557post runmode documentation in LVT


emkemp:
Need to add 557post runmode documentation in LVT.

This issue was first raised with the LIS 7.2 deliverable to 557WW, but has yet to be fixed.

PR #157, opened 2019-06-19
Fix issue #156 for 557WW

emkemp:
Resolves issue #156


emkemp:
I'm assigning this to Jim since I cannot convert lvt.config.master to LaTeX.

Issue #161, opened 2019-06-24
Add strict check with warnings configure option for LDT


emkemp:
Add a "strict check with warnings" configure option for LDT.  Will mimic option for LIS.

Intent is to allow the ifort compiler to automatically create interfaces, and thereby identify mismatches in argument lists between subroutine calls and signatures.

PR #162, opened 2019-06-24
Added "strict checks with warnings" option for LDT

emkemp:
Borrows "strict checks with warnings" logic from lis/arch/Config.pl.

Resolves:  #161


dmocko:
Compiles with "strict checks" optimization level -3 - but only if the code described in issue #166 is not compiled.  That will have to be addressed through issue #166.

Approving this pull request.

Issue #163, opened 2019-06-24
Add strict checks with warnings configure option for LVT


emkemp:
Add a "strict check with warnings" configure option for LVT. Will mimic option for LIS.

Intent is to allow the ifort compiler to automatically create interfaces, and thereby identify mismatches in argument lists between subroutine calls and signatures.

PR #164, opened 2019-06-24
Added "strict check with warnings" option for LVT

emkemp:
Borrows "strict checks with warnings" logic from lis/arch/Config.pl.

Resolves: #163


dmocko:
Compiles with "strict checks" optimization level -3

PR #183, opened 2019-06-27
Adds the ability to process the OPT/UE outputs in LDT

### LINKED ISSUE #182: Add the capability in LDT to generate calibrated model parameter maps from LIS OPT/UE outputs

sujayvkumar:
Enables LDT to read the OPT/UE outputs from LIS and extract the calibrated
parameters. A new runmode to support this feature has been added ('OPTUE parameter processing')
In this runmode, the LDT will extract the parameters from the OPT/UE output file
and then add them to the LIS domain/parameter file generated by LDT.

Resolves #182


sujayvkumar:
Example testcase at: /discover/nobackup/projects/ifloods/svkumar/SUSMAP/procOPTUE

PR #194, opened 2019-07-08
Feature/jules50 snodep da

yyoon4:
Add SNODEP DA support to Jules 5.0

Here is the link for the testcase:
/discover/nobackup/yyoon4/lis7/Testcase/Jules.5.x/Jules.5.0/EnKF

Please let me know if you have any questions.

### DIFF
--- a/lis/make/default.cfg
+++ b/lis/make/default.cfg
@@ -1268,8 +1268,10 @@ path: RESTRICTED/jules/jules.5.0/jules.5.0,
       RESTRICTED/jules/jules.5.0/jules.5.0/src/util/grid,
       RESTRICTED/jules/jules.5.0/jules.5.0/src/util/cube
 dependent_comps: virtual_da,
+                 virtual_da_obs_snodep,
                  virtual_routing
 virtual_da path: RESTRICTED/jules/jules.5.0/jules.5.0/da_soilm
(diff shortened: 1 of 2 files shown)


dmocko:
Eric,

It is frustrating that it took a while to identify the problem with the default.cfg.  I do overall like the new structure, but it can be difficult to find issues like this sometimes.

I believe that long-term we don't want to add JULES as "true" to the default.cfg file, as the LISF code should compile without access to any JULES code.

David

yyoon4:
Sorry for the inconvenience. 

To run,
You don't need to change the "default.cfg" file. In stead of the file, you need change the "user.cfg" file like this:

JULES.5.0: On
JULES.5.0_DEV: On

Yeosang

emkemp:
I'm testing this again with the restricted JULES code.  Will let you know how it goes.

yyoon4:
I reproduced LIS domain and parameter data file. Elevation field looks OK now. 
You don't need to change anything in lis.config, but just in case:
/discover/nobackup/yyoon4/lis7/Testcase/Jules.5.x/Jules.5.0/proc_LSM/lis_input.10k.us.ps41.nc

yyoon4:
Update LIS-JULES code to address negative snow and soil moisture variables (from @shugongwang)

Issue #200, opened 2019-07-12
Configuring LDT little_endian causes problems processing TBOT


dmocko:
If a user chooses the "little_endian" option when configuring LDT, then compiles and runs - LDT does not correctly process the TBOT bottom soil temperature data.

This dataset is natively "big_endian", and the results from LDT for this field will be incorrect if LDT had been configured with "little_endian".

Issue #199, opened 2019-07-11
Configuring LDT little_endian causes problems processing GTOPO and SRTM


dmocko:
If a user chooses the "little_endian" option when configuring LDT, then compiles and runs - LDT does not correctly process the GTOPO and SRTM topography data.

These datasets are natively "big_endian", and the results from LDT for these fields will be incorrect if LDT had been configured with "little_endian".

PR #201, opened 2019-07-12
Fix to LDT little_endian issues for GTOPO, SRTM, and TBOT

dmocko:
This commit fixes issues relating to configuring LDT with
little_endian, but trying to read the natively big_endian
datasets from GTOPO, SRTM, and TBOT.

The primary fixes are to the "???_Native" options for the
topography and to the "ISLCP1" option for TBOT, as these
native datasets are big_endian.

However, related fixes were also made to the "???_LIS"
options for datasets provided by the LIS team.  These
datasets are legacy data, and have traditionally been
produced in big_endian.  Users should consider moving
to the "???_Native" datasets for these parameters.

Resolves: #199
Resolves: #200


dmocko:
This bug fix could also be merged into the support/lisf-557ww-7.3 branch.

PR #211, opened 2019-07-25
Enables the support for SMAP L2 soil moisture retrievals

### LINKED ISSUE #208: Add support for SMAP level 2 soil moisture retrievals in LVT
Currently plugin only supports L3 soil moisture retrievals

sujayvkumar:
The SMAP soil moisture plugin in LVT has been extended to support
the L2 flavors of SMAP data. The L2 standard retrievals at 36km
and the enhanced version at 9km are supported.

Resolves #208


sujayvkumar:
An example test case can be found at : /discover/nobackup/svkumar/LISF_testing/208

Issue #214, opened 2019-08-01
SNODAS LVT code gives wrong and/or incomplete messages to log file


dmocko:
The SNODAS code in LVT gives wrong information to the log file:

------
 [INFO] Reading SNODAS SWE data 
 ./input/STN_DATA/SNODAS/20031003/us_ssmv11034tS__T0001TTNATS2003100305HP001.dat
 [INFO] Successfully processed SWE from 
 ./input/STN_DATA/SNODAS/20031002/us_ssmv11036tS__T0001TTNATS2003100205HP001.dat
 [INFO] Reading SNODAS SNWD data 
 ./input/STN_DATA/SNODAS/20031003/us_ssmv11036tS__T0001TTNATS2003100305HP001.dat
 [INFO] Successfully processed SNWD from 
 ./input/STN_DATA/SNODAS/20031003/us_ssmv11036tS__T0001TTNATS2003100305HP001.dat
------

The problem is the second message ("Successfully processed SWE") which points to the SnowDepth file (not the SWE file) and for the previous date.  From looking at the code, it appears that only the log message needs to be corrected, and the SWE and proper date are correctly being read in and processed by LVT.

Additionally, no log message is given if LVT tries to read a SNODAS file, but it is not found.  This lack of a message makes it difficult for a user to diagnose any issues, such as not pointing to the correct path for the data in the lvt.config file.

PR #215, opened 2019-08-01
Fix SNODAS LVT log messages and big_endian data read

dmocko:
This pull request fixes two issues with log messages
in the SNODAS LVT reader.  The first is that the line
showing successful processing of the SWE dat file now
prints the correct file name.  The other is that LVT
will print a log message when it cannot find a SNODAS
file for which it is looking.

Also addressed in this pull request is to specify in
the open statement that the SNODAS dat binary files
are natively "big_endian".  Now LVT will produce the
same results when processing SNODAS when configured/
compiled with either big or little endian.

Resolves: #214

### DIFF
--- a/lvt/datastreams/SNODAS/readSNODASObs.F90
+++ b/lvt/datastreams/SNODAS/readSNODASObs.F90
@@ -20,22 +20,22 @@ subroutine readSNODASObs(source)
   implicit none
 !
 ! !INPUT PARAMETERS: 
-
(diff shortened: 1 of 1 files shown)


dmocko:
Sorry - had a typo in my original commit message.
This addresses #214 - not 124.

Testcase input/output data is here:
    /discover/nobackup/projects/lis/tools/pull_requests/215

Issue #202, opened 2019-07-12
Inconsistent LDT/LIS/LVT code and incomplete documentation for the HyMAP router


dmocko:
The HyMAP and HyMAP2 streamflow routers in LIS have some small inconsistencies in the parameters, variable names, and units between the LDT, LIS, and LVT packages.

Also, HyMAP parameters are natively in big_endian format, so LDT needs to be modified so the code will run as expected even if LDT was configured with little_endian.

Finally, the documentation for the Users' Guide and associated files needs to be updated, primarily for the HyMAP2 router.

PR #217, opened 2019-08-02
Fix HyMAP/HyMAP-2 documentation and make codes more consistent

dmocko:
This pull request adds documentation and clarifies details about
the HyMAP and HyMAP-2 routing models and required parameters to
the LDT and LIS Users' Guides.  It also makes the codes more
consistent between LDT, LIS, and LVT, with respect to required
parameters, units, and variable names.

Also included:

- LDT now hard-codes the open statements to read HyMAP parameter
  data in "big_endian" format.  These datasets are typically in
  "big_endian", and now LDT will produce the same results when
  reading these parameters when configured/compiled with either
  big or little endian.

- These parameters for the HyMAP-2 routing model are optional,
  and no longer required to be processed by LDT if they will
  not be used in the LIS/HyMAP-2 simulation: a) river flow type
  map; b) baseblow dwi ratio map; and c) runoff dwi ratio map.

- lis/configs/MODEL_OUTPUT_LIST.TBL.adoc was updated to correct
  units of HyMAP output variables, and add additional variables.

Resolves: #202


dmocko:
Since this pull request involves changes for consistency between LDT, LIS, and LVT, I have prepared testcase input/output for all of them "end-to-end".  For convenience, however, I have also included copies of the LDT lis_input file in the LIS and LVT testcases, as well copies of the LIS output in the LVT testcase.

The README and testcase files are found here:
/discover/nobackup/projects/lis/tools/pull_requests/217

emkemp:
I'm unable to reproduce the LIS routing results for HyMAP2.  All of the variables are filled with zeros or missing values.  I tested with the Intel compilers with both big and little endian.  Not sure what the problem is.

dmocko:
Did you run with multiple processors?  I believe that HyMAP-2 currently does not run parallel, and will return zero/missing values.  This is true even for the code before the pull request.

If you ran on a single processor, and are getting zero/missing values, I will resolve it.

karsenau:
Did Sujay and Augusto already implement HyMAP2 in parallel as of earlier this year?

emkemp:
@dmocko I did run HyMAP2 with multiple processors (63x4), as those are the settings in the lis.config.HyMAP2 test file.  Is that in error?

@karsenau I will have to check.

dmocko:
@karsenau I believe it was announced as running in parallel earlier this year, but I have not done it.  We may need to open this as a separate Issue.

@emkemp Sorry - I think I just had those processor settings from a previous run, but had always actually run this testcase using a single processor.

emkemp:
@dmocko I reran LIS with a single processor, and I reproduced your HyMAP2 results.  LDT results were reproduced earlier.  I will test LVT now.

emkemp:
I reproduced the LVT results.  Also, I reviewed the code and tentatively approve.

One gotcha:  There is a conflict with the ldt.config_master file.  I think I can resolve this via GitHub's web editor, but I want to confer with Jim before proceeding.  Hopefully this pull request can be resolved and merged in tomorrow morning.

PR #219, opened 2019-08-03
Enables the support for LPRM VOD in LVT

### LINKED ISSUE #218: Add support for LPRM VOD retrievals in LVT

sujayvkumar:
This commit enables the support for processing VOD retrievals from
the Land Parameter Retrieval Model. The plugin supports the processing
of VOD retrievals from the X- and C-band sensors.

Resolves #218


sujayvkumar:
An example test case can be found at : 
/discover/nobackup/svkumar/LISF_testing/218

sujayvkumar:
I've updated the test case so that there are no 'missing' metric values. Also restructured the directory so that the testcase is self consistent (and not pointing to the disk). Please see : 
/discover/nobackup/svkumar/LISF_testing/218

It's not super clear which directories need to be tarred for a LVT only case (and sorry, I am not going to bother with the grads files - I think we need to get away from them).

emkemp:
Hi Sujay:

I do not have read permissions for ./input or ./STATS in your latest test case.  I cannot proceed with this pull request until I have access.

sujayvkumar:
Changed

emkemp:
Okay, test case runs w/o issue with both Intel and GCC compilers.  I already gave the thumbs up for code and documentation.

I suggest opening a separate pull request for the LVT test case, as I don't know when Jim will get around to incorporating it into his test suite.  In the mean time, I'm merging the code into master.

Issue #213, opened 2019-07-30
Improve LDT - HyMAP input restart warning messages


karsenau:
Currently LDT does not inform/warn the user that an input HyMAP restart (binary) file is present or not when performing upscaling/downscaling of restart files, which results in traceback of error messages.

Improve messaging for users.

dmocko:
Closing issue, which was resolved via pull request #223

PR #223, opened 2019-08-14
Improve LDT log messages for HyMAP restart files

karsenau:
Added better log messages to inform and warn user
that an input HyMAP restart (binary) file is present or
not when performing upscaling/upscaling of restart files.

### DIFF
--- a/ldt/core/LDT_ensRstMod.F90
+++ b/ldt/core/LDT_ensRstMod.F90
@@ -365,19 +365,27 @@ module LDT_ensRstMod
             ftn = LDT_getNextUnitNumber()
             ftn2 = LDT_getNextUnitNumber()
 
-            open(ftn,file=LDT_rc%inputrst,status='old',&
-                 form='unformatted',access='sequential',iostat=ios)
+            ! Check if input restart file is present:
+            inquire( file=trim(LDT_rc%inputrst), exist=file_exists )
+            if( file_exists ) then
+               write(LDT_logunit,*) "[INFO] Opening HYMAP input restart file, "
+               write(LDT_logunit,*)  trim(LDT_rc%inputrst)
+               open(ftn,file=LDT_rc%inputrst,status='old',&
+                    form='unformatted',access='sequential',iostat=ios)
+            else
+               write(LDT_logunit,*) "[ERR] HYMAP input binary restart file, "
+               write(LDT_logunit,*)  trim(LDT_rc%inputrst)
+               write(LDT_logunit,*) " is missing. Stopping run ..."
+               call LDT_endrun
+            endif
 
- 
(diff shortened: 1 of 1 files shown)


karsenau:
This pull request address issue #213 

https://github.com/NASA-LIS/LISF/issues/213

dmocko:
Kristi,

I've reviewed this pull request, and generally it seems OK.  Before I merge it, can you please explain why you changed the code from stopping if the HyMAP restart file would be overwritten to just making it a warning, if it would be overwritten?  Also, I noticed that you made this change only in the "upscale" routine, but not in the "downscale" routine.  Can you please explain why?

I'm sure there's a great explanation, but I'm failing to understand it.

If this is indeed the code that is needed, I do have one very small commit request to edit the comment on line 440 in the new code, to also indicate that a warning will be given, instead of preventing an overwrite.

Thanks,

David

karsenau:
Thank you for the excellent review and questions.

The code was changed from stopping and not overwriting the file to warning and overwriting to be more consistent with the LSM restart file upscaling process and for improving automation process (when used in routine running scripts). Also, the co

For the downscaling code, it was an oversight about not being updated similarly there.  I can update the code and make another pull request to make it more consistent as well.

karsenau:
I have made updates to the downscaling LIS HyMAP ensemble restart writing part of the code, as per your questions and comments.

Please let me know if you have any other questions or concerns.
Thanks!

ldt/core/LDT_ensRstMod.F90
```diff
@@ -413,21 +421,29 @@ subroutine upscale_ensembleRst()
...
+               !     form='unformatted',access='sequential',iostat=ios)
+            else
+               write(LDT_logunit,*) "[ERR] HYMAP input binary restart file, "
+               write(LDT_logunit,*)  trim(LDT_rc%inputrst)
+               write(LDT_logunit,*) " is missing. Stopping run ..."
+               call LDT_endrun
+            endif
 
- 
             ! If output ensemble restart file exists, don't overwrite ...
```
dmocko: Can you please edit the comment on this line (and commit it) to indicate that a warning will be provided, instead of preventing the file from being over-written?

karsenau:
Ok, made that additional comment update.

ldt/core/LDT_ensRstMod.F90
karsenau: Ok, I just updated that comment line. Thanks.

dmocko:
Thanks for the edits.  Code and comments are consistent.  I approve this pull request.

dmocko:
Sorry - hit the "close" button by mistake.  I re-opened and then merged the commit, which closes this pull request.

PR #225, opened 2019-08-15
Fix/smapreader

LIS-navari:
The LDT reader for SMAP soil moisture L3 was modified to read the latest version of SMAP data based on the Composite Release ID (CRID) number. The reader asks a user to enter the part of Composite Release ID (i.e., a three-character string like R16) 

Test case: /discover/nobackup/mnavari/my_projects/op/US_10km/new_smap_reader_ldt


emkemp:
Closed/rejected since this is supposed to go into support/lisf-557ww-7.3 instead of master.

PR #226, opened 2019-08-15
LDT SMAP reader modified to asks user for Composite Release ID (CRID)…

LIS-navari:
… number.

The reader reads the latest version of data based on user input.


LIS-navari:
Test case : /discover/nobackup/mnavari/my_projects/op/US_10km/new_smap_reader_ldt

PR #253, opened 2019-09-13
Update ldt.config testcase config files

jvgeiger:


jvgeiger:
Running LDT with our internal acceptance test case config files showed that a few of those config files must be updated.

PR #362, opened 2019-09-27
Updates the synthetic soil moisture reader to use NetCDF files

### LINKED ISSUE #361: Update the synthetic soil moisture DA reader to use the simulated 'observations data' from LDT
Current reader relies on using binary data which need to be generated from a stand alone program (not available within LIS)

sujayvkumar:
Updated the reader that processes synthetic soil moisture observations
for DA to use NetCDF files generated by the OSSE capabilities in LDT.
The existing version relies on binary files which need to be created
outside of LIS by the user.

Fixes #361


sujayvkumar:
Approving this pull request myself, since this is a minor code change overall.

PR #364, opened 2019-09-30
Allows ensemble averaging of LIS ensemble outputs

### LINKED ISSUE #363: ensemble averaging is not enabled in LVT for LIS output
If the LIS output is in "2d ensemble grid space", then LVT can read them and output each ensemble members, as long as the 'Number of ensembles in the LVT analysis:' option to set to the number of ensembles used in the LIS run. 

If the 'Number of ensembles in the LVT analysis: ' is set to 1, then LVT (currently) does not average the LIS ensemble.

sujayvkumar:
The bug fix allows the calculation of ensemble mean values from the LIS
ensemble

Fixes #363

### DIFF
--- a/lvt/core/LVT_LISoutputHandlerMod.F90
+++ b/lvt/core/LVT_LISoutputHandlerMod.F90
@@ -10268,6 +10268,8 @@ subroutine get_moc_attributes(modelSpecConfig, head_dataEntry, &
   subroutine mapLISdataToLVT(source,lvtdataEntry, lisdataEntry)
 ! !ARGUMENTS: 
 
+    implicit none
+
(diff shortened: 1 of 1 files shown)


sujayvkumar:
A sample testcase can be found at : /discover/nobackup/svkumar/LISF_testing/363

emkemp:
DISREGARD PREVIOUS COMMENT

I had ncview running with bilinear interpolation turned on.  When I turn that off (turn raster  mode on), the pattern goes away and I see values for the 4 grid points.

emkemp:
Further test with the GNU compilers also look good.  Code looks good.  I approve.

PR #366, opened 2019-10-04
Feature/jules5x

shugongwang:
JULES (5.x) has been registered in the plugins/LIS_lsmrouting_pluginMod.F90 for enabling HyMap routing. The code has been tested with LIS-JULES run in the NLDAS domain.


dmocko:
Did you mean to just edit/commit the one file?  Looks like every file was changed.  Also, standard practice is to first open an "Issue" and then attach the pull request to it.

Finally, the team needs to do better with pull request and issue titles.  "Feature/jules5x" does not describe the problem.  If it's just the one file, then something like this would be better:

Issue title: JULES-5.x LSM in LIS does not work with routing
Pull request title: Add routing to the JULES-5.x LSM in LIS.

Noted that we should always put in LDT, LIS, or LVT in the title somewhere as well.

emkemp:
Hi Shugong:

You will need to provide us with a test case so we can confirm the code in this pull request works correctly.

Also, you will need to make the actual JULES LSM code available to us on Discover (NOT ON GITHUB) so we can run the test case.

emkemp:
Shugong:  Please refer to this document when assembling your test case.
https://github.com/NASA-LIS/LISF/blob/master/docs/howto_create_lis_testcases/howto_create_lis_testcases.adoc

shugongwang:
Hello,


Here is my test case: /discover/nobackup/swang9/Models/JULES/NLDAS.5.5/OL_GL8


Here is my LIS directory: /discover/nobackup/swang9/LIS/LISF

Thanks and have a good weekend!

Shugong

shugongwang:
Hi Eric,

I can certainly make the test case according to the guideline. However, my question is that the code change is not a major update. Instead, I just registered JULES 5.x in the LSM routing plugin file. For such a minor change, do we still need to make such a test case? My concern it time spending.

Thanks,

Shugong

dmocko:
Again, if the only change is registering the JULES 5.x LSM in the LSM routing plugin file, why does this pull request have over 4,000 files changed?

shugongwang:
Good question but I have no idea. There might be wrong on my master branch. I worked on it when I was working on JULES 5.x container. I may have to get my master branch fixed first. Any suggestions?

Thanks,

Shugong

dmocko:
It looks like the vast majority of the changed files are trying to change permissions from 644 to 755, with the actual contents of the files unchanged.  This might be related to a setting you have in your LISF/.git/config file.  I always change this setting from the default of "true" to "false".

[core]
.....
        filemode = false

We don't want to change the permissions on the files (especially just about all of them), so we shouldn't approve this pull request at all right now.  You should work with @jvgeiger to clean all of this up.

Aside from that, looking through the files that are actually changed, it looks like you also commit-ed and pull request-ed the JULES 5.x container code.  Did you want to do that now, or only change the LSM routing plugin file?  Again, @jvgeiger can help you clean this up.

About the test case - yes, it's only a minor code change, but it is a new feature, so it should have a packaged testcase associated with it, so the approvers can confirm and approve the pull request.

shugongwang:
Thanks a lot David for your detailed help! I will work with Jim at his convenience in next week to solve this problem.

Shugong

jvgeiger:
The problem with file permissions was caused by commit c39610b "test".

jvgeiger:
Shugong is rebuilding this pull request.

PR #372, opened 2019-10-08
Update LIS config documentation for MERRA2 reader

dpsarmie:
This commit adds documentation for an option to use 2m wind
fields when using the MERRA2 reader. The documentation was
not originally included when the commit was first made. This
option was meant to compliment a secondary reference
evapotranspiration calculation option that is available in
LVT.

There was no issue opened for this fix.


dmocko:
Looks good.  Always glad to keep documentation updated.  Will approve and merge.

PR #379, opened 2019-10-16
"divided by zero" issue of Noah3.9 has been fixed.

yhkwon81:


jvgeiger:
That sounds like a good idea since we are touching the original physics.

PR #382, opened 2019-10-17
Remove redundant assignment to precip_regrid

jvgeiger:
precip_regrid was being set to LIS_rc%udef before it was allocated.
This caused LIS, compiled with the GNU compilers (version 7.3), to hang.
Further down in the routine, precip_regrid is allocated and then set
to LIS_rc%udef.

Resolves: #381


dmocko:
I have reviewed the code, and agree with the change.  I will merge this pull request.

PR #391, opened 2019-10-29
Fix read restart

jvgeiger:
This fixes the insufficient memory error that Yonghwan was getting when running NoahMP 4.0.1.

### DIFF
--- a/lis/core/LIS_historyMod.F90
+++ b/lis/core/LIS_historyMod.F90
@@ -4606,6 +4606,7 @@ contains
     integer :: varid
     integer :: status
     integer :: count1 ,c,r,npatch,t,gid,stid,tid
+    integer :: glbnpatch_size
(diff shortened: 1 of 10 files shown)


jvgeiger:
Sorry, I had already upload my test case files to the dataportal.  You want

noah36restarttest_input_v72.tar.gz
noah36restarttest_output_v72-intel17.tar.gz
noah36test_ldt_v72.tar.gz

in /discover/nobackup/jvgeiger

Sorry, I don't see 2017 anywhere in testcases/surfacemodels/land/noah.3.6_restart.

emkemp:
Hi Jim:

I exactly reproduce results when using Intel compilers.  The GNU versus target output results are messier and may warrant more investigation, but I suspect the cause has nothing to do with this change.  I therefore approve.

Issue #389, opened 2019-10-29
LIS Noah-MP-4.0.1 missing some output variables/units


dmocko:
The Noah-MP-4.0.1 LSM main routine in LIS does not provide some required output variables, or provide output variables in required units.

dmocko:
Closing this issue as the code has been merged

PR #393, opened 2019-10-29
Add missing output variables/units to LIS Noah-MP-4.0.1 LSM

dmocko:
This commit adds some missing output variables or units to the
output of the Noah-MP-4.0.1 LSM in LIS.  Added are the option
to output Qs/Qsb in units of [kg m-2], output Albedo in units
of [%], and the new output of RelSMC (relative soil moisture).

The RelSMC code was borrowed from its calculation in Noah-3.X
LSMs within LIS.

Resolves: #389


dmocko:
Testcase found in:

/discover/nobackup/projects/lis/tools/pull_requests/390

PR #396, opened 2019-10-31
Fix RHMin calculation in Noah 3.9; add to NoahMP 4.0.1

emkemp:
Changed code to reset minimum air temperature threshold for RHMin every 3 hours in Noah 3.9.  Added RHMin to NoahMP 4.0.1.

This will allow RHMin to be calculated every three hours even if a long-term simulation is made.  Previously the temperature threshold (and RHMin) was only updated when a colder air temperature was encountered, which could lead to wrong answers.

### DIFF
--- a/lis/core/LIS_historyMod.F90
+++ b/lis/core/LIS_historyMod.F90
@@ -483,6 +483,8 @@ contains
 
     call LIS_rescaleCount(n,group_temp)
 
+    write(LIS_logunit,*)'[INFO] Writing to ',trim(lsmoutfile) ! EMK
+
     if(LIS_rc%wout.eq."binary") then 
        if(LIS_masterproc) then 
           open(ftn,file=lsmoutfile,form='unformatted')
--- a/lis/core/LIS_historyMod.F90
+++ b/lis/core/LIS_historyMod.F90
@@ -483,7 +483,8 @@ contains
 
     call LIS_rescaleCount(n,group_temp)
 
-    write(LIS_logunit,*)'[INFO] Writing to ',trim(lsmoutfile) ! EMK
+    write(LIS_logunit,*)'[INFO] Writing surface model output to:  ', &
+         trim(lsmoutfile) ! EMK
 
     if(LIS_rc%wout.eq."binary") then 
        if(LIS_masterproc) then 
--- a/lis/surfacemodels/land/noah.3.9/noah39_lsmMod.F90
+++ b/lis/surfacemodels/land/noah.3.9/noah39_lsmMod.F90
@@ -259,7 +259,7 @@ contains
 
        ! Initialize min/max values to implausible values.
        noah39_struc(n)%noah(:)%tair_agl_min = 999.0
-       noah39_struc(n)%noah(:)%tair_agl_min = 999.0
+       noah39_struc(n)%noah(:)%rhmin = 999.0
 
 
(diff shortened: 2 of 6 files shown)


emkemp:
Hi David:

I improved (?) the log message with my latest commit--it should be part of the overall pull request.

Regarding the reset interval--it's my understanding that this RHMin product is defined as no more than 24-hr values, so I'm not sure if setting this to the surface model output interval makes sense.  (After all, what if history dumps are reduced to weekly for a long run?)  At the same time, RHMin more frequently than 3 hours is not desired by the customers.  This argues for using a separate time interval, and setting that interval to 3 hours.

An alternative is to add the interval as a new lis.config option.  But this seems overkill to me.

dmocko:
> Hi David:
> 
> I improved (?) the log message with my latest commit--it should be part of the overall pull request.
> 
> Regarding the reset interval--it's my understanding that this RHMin product is defined as no more than 24-hr values, so I'm not sure if setting this to the surface model output interval makes sense. (After all, what if history dumps are reduced to weekly for a long run?) At the same time, RHMin more frequently than 3 hours is not desired by the customers. This argues for using a separate time interval, and setting that interval to 3 hours.
> 
> An alternative is to add the interval as a new lis.config option. But this seems overkill to me.

Hmmm - because this is a specific requirement for a particular customer right now, then your methodology on the reset interval is probably sufficient.  I also agree that we can probably avoid adding this as a lis.config option.

The only other thing I will add is that the Tair max/min (or the max/min for any variable as set in the MODEL OUTPUT TBL) is also directly tied to the output interval.  In your example, when writing the history weekly, the Tair max/min in that week's history file will also be the maximum and minimum of the Tair over that entire week.  Why shouldn't RHMin be the same?

David

emkemp:
Hi David:

Re timing interval of Tair min... that's a very good point.  And LVT uses the Tair min to find the appropriate 24hr RHMin from 8 sequential 3-hr outputs.

On reflection, it's probably best for RHMin to use the same output interval as Tair min.  Perhaps an error message should be logged if RHMin is selected but the output interval is more than 24 hours?

dmocko:
Probably give it a [WARN] to the log file, if the RHMin is selected and the output interval is more than 24 hours.

I will wait for additional commits.  Thanks.

emkemp:
Okay, I just pushed an update for Noah3.9 and NoahMP4.0.1.  Let me know what you think.

dmocko:
Lines 261 and 262 of:
    lis/surfacemodels/land/noah.3.9/noah39_lsmMod.F90
are identical.  I believe line 262 should be %rhmin instead.

In the "main" routines for both Noah-3.9 and Noah-MP-4.0.1, the %tagl_agl_min is reset to 999, while %rhmin is not.  But they are both initialized to 999.  Should %rhmin also be reset to 999, to be consistent with the lsm_ini routine?

emkemp:
Oops.  I fixed the repeat initialization in Noah3.9.

The rhmin value should *not* be reset at the bottom of the _main routine, so the stored value can be written to the history file.  It will instead be overwritten on the next time step.

dmocko:
OK - I approve these changes.  Let me test before I approve the pull request.

dmocko:
I have tested this successfully for both Noah-3.9 and Noah-MP-4.0.1, and the code is ready to be merged.

Issue #402, opened 2019-11-05
JULES 5.0 DA is not compiled by default in LIS


dmocko:
JULES 5.0 data assimilation plugins are not compiled by default in LIS.

The default/user.cfg files need updating, as well as code in the plugins directory.

PR #403, opened 2019-11-05
Compile JULES DA plugins when compiling LIS with JULES LSMs

dmocko:
This commit enables by default the compilation of JULES DA plugins
when compiling LIS with a JULES LSM.  Previously, a separate compiler
flag "JULES_5_0_DEV" would need to be defined in the user.cfg file.
This commit removes this flag entirely from the code, leaving the
JULES DA plugins to be compiled only when LIS is compiled with a
particular version of the JULES LSM.

Resolves: #402


emkemp:
Jim wants to review this himself, so I will not merge this now.

jvgeiger:
I approve.

PR #412, opened 2019-11-18
More fixes to creating output directories

emkemp:
(1) Make sure only master process creates directory
(2) Simplify check of stat function call

### DIFF
--- a/lis/core/LIS_create_subdirs.c
+++ b/lis/core/LIS_create_subdirs.c
@@ -130,27 +130,14 @@ int FTN(lis_create_subdirs) (int  *length,
                     error = 1;
                 }
 
-            } else {
-                
-                /* If stat was successful, check if path is a directory.  Use
-                   POSIX S_ISDIR macro */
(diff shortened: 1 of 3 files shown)


dmocko:
In lis/core/LIS_readConfig.F90, should we print an [ERR] log message when ios is not equal to zero?

Note we didn't log an error message before this pull request, but should we now take the opportunity to add one?

I'm referring to Line 105.

emkemp:
I added the requested code.

dmocko:
Thanks.  Instead of just "flush", shouldn't we "call LIS_flush"?

emkemp:
I added calls to the LIS_flush subroutine.

(I will add that that LIS_flush is unnecessary in Fortran 2003, which now has the intrinsic library function 'flush'.  But this is not the time to remove LIS_flush from the LIS code.)

dmocko:
Code looks good and runs fine on Discover.  If Jim's testing also looks good, then we can approve.

PR #414, opened 2019-11-20
CRITICAL BUG FIX:  Added "-" units for RelSMC in LVT

emkemp:
Added "-" units for RelSMC in LVT.

This is an operational requirement for AFW.

### DIFF
--- a/lvt/core/LVT_LISoutputHandlerMod.F90
+++ b/lvt/core/LVT_LISoutputHandlerMod.F90
@@ -3158,7 +3158,7 @@ contains
                   3, nsize, nensem, (/"%    ", "m3/m3", "-    "/), &
                   1, (/"-"/),&
                   valid_min=(/0.0, 0.0, 0.0/), &
-                  valid_max=(/1.0, 1.0, 1.0/), &
+                  valid_max=(/100.0, 1.0, 1.0/), &
                   gribSFC=grib_depthlvl, gribLvl=0)     
           endif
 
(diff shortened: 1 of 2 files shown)


emkemp:
PLEASE HOLD OFF ON ACCEPTING THIS COMMIT

I forgot to update the valid max and min for the extra unit option.  But I cannot work on this right now--maybe Thursday afternoon.

dmocko:
A few requests:

1) I'd prefer to see the old code removed, instead of commented out.

2) valid_min and valid_max will also need to be updated for RELSMC.  The first valid_max value should be 100.0 (for percentage) - and we need a 3rd value for the min and max on this line.  It can also be 0.0/1.0, as it is for m3/m3.

3) Don't we also need to add units of [-] to RelSMC in LVT_histDataMod.F90?

David

emkemp:
Pull request still  not ready yet.  Please stand by for further updates.

emkemp:
I believe this pull request is ready for review and (hopeful) acceptance.

dmocko:
Looks good to me.  One small request.  Instead of changing [ERR] to [ERR1], [ERR2], etc. - can you change it so the message says something similar to: '[ERR] binary output routine to diagnose....' ?

emkemp:
Whoops!  Those changes were for debugging.  I'll fix them.

dmocko:
Thanks for all the modifications.  Code is ready for merging.

PR #418, opened 2019-11-22
Updated RHMin calculation in NoahMP.4.0.1.

emkemp:
The earlier version used the CALHUM subroutine in NoahMP.4.0.1 to calculate
saturation specific humidity, following similar logic in Noah.3.6.  However,
the NoahMP version actually calculates saturation mixing ratio.  This
led to a different (and physically inconsistent) value of RHMin compared
to the values produced by Noah.3.9 and JULES.5.0 with identical atmospheric
forcing.

The new code takes the same approach as JULES.5.0, with necessary calculations
done in-line in NoahMP401_main but with formula matching the Noah.3.6 code.

### DIFF
--- a/lis/surfacemodels/land/noahmp.4.0.1/NoahMP401_main.F90
+++ b/lis/surfacemodels/land/noahmp.4.0.1/NoahMP401_main.F90
@@ -27,7 +27,6 @@ subroutine NoahMP401_main(n)
     use LIS_logMod, only     : LIS_logunit, LIS_endrun
     use LIS_FORC_AttributesMod
     use NoahMP401_lsmMod
-    use module_sf_noahmplsm_401, only: calhum, noahmp_parameters
 
     implicit none
 ! !ARGUMENTS:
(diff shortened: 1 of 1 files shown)


emkemp:
NOTE:  The code in this pull request has only been compiled.  I need David's short LIS73-NoahMP401 test to be rerun with this code to ensure RHMin matches the other two LSM runs.

dmocko:
Re-running the Noah-MP-4.0.1 test run with this latest pull request....

emkemp:
I postprocessed David's output with this patch, and I confirm the RHMin values are now the same between Noah 3.9, NoahMP 4.0.1, and Jules 5.0.  I specifically looked at three-hourly output from 03Z 2 Nov to 00Z 3 Nov 2007, with visual inspection of ncview plots.

I believe the patch works and is functionally ready to merge in.

dmocko:
I also directly compared the LIS output, and found that the Noah-MP-4.0.1 RHMin fields are now the same as Noah-3.9 and JULES-5.0.  This is ready to merge.

PR #423, opened 2019-12-03
Changed NoahMP soil thickness units to centimeters in history output.

emkemp:
This change is for consistency with other LSMs, and for LVT (in 557 post mode).
Note that the units in the internal variable (used by NoahMP itself) remain in
meters.


emkemp:
David produced a short test run with this patch, and I just created GRIB2 files with LVT from that run.  Bottom line is the soil thicknesses now look correct.

I believe that this patch (at least functionally) is ready to merge in.

dmocko:
I agree - will merge in this code.

PR #425, opened 2019-12-03
Added Greenness_inst for Noah and NoahMP.

emkemp:
This variable is included in GRIB2 postprocessing but was erroneously omitted
for netCDF postprocessing.

Note that Greenness is not used with JULES, so omitting it from the JULES
variable list is intentional.


dmocko:
Looks good, thanks.  Will merge this in now.

PR #426, opened 2019-12-04
Remove string "version 7.2r" from configure and compile messages

jvgeiger:


emkemp:
Code looks good.  Run tests show nothing was broken.  I approve.

PR #434, opened 2019-12-12
Update netcdf references to recent LDT paper

karsenau:
Included the LDT description paper citation within
 routines that generate any netcdf files that contain
 references for LDT and LIS as global attributes.


emkemp:
Code compiles w/o issue.  Since this just changes metadata in the netCDF output, I don't see the need to manually run this.  I'll merge the changes now.

PR #437, opened 2019-12-13
Replaces "LDTSI" with "USAFSI" in LVT

emkemp:
Also adds runtime option to specify USAFSI filename prefix in lvt.config.  (This allows use of older snow and ice analyses with the now-retired 'ldtsi' prefix.)


emkemp:
PLEASE WAIT ON PROCESSING THIS PULL REQUEST

I need to do an end-to-end USAFSI run with all the changes to LDT, LIS, and LVT.

PR #438, opened 2019-12-14
Fix several errors raised when compiling with strict checks enabled

jvgeiger:


emkemp:
Code looks good, and compiles with both Intel and GNU compilers.  I approve.

PR #439, opened 2019-12-16
Added Elevation to exception list for smoothing.

emkemp:


emkemp:
NOTE:  Additional variables could be added to the Exception list, either in this pull request or in a new one.  But I have not received any feedback on variables other than Elevation, so I'm submitting this pull request now.

PR #440, opened 2019-12-17
Fix for domains crossing PM in LDT

karsenau:
With the latest LDT updates made to the interp routines,
  one more fix is required to compute_earth_coord_latlon.F90
  to address an issue related to crossing the Prime Meridian (PM).

 This should be the last fix committed for this Git fix/410
  issue.

 For the other routine committed, I simply changed some write statements
  to write to the LDT log unit file.

For testing, I tested over three different domains throughout the globe and for both latlon and lambert projections. This latest commit fixes an issue and worked for all the tests run.


emkemp:
Code compiles with Intel and GCC.  Code looks fine.  I approve.

PR #441, opened 2019-12-18
557 post mode now uses Welford algorithm for standard deviation.

emkemp:
The previous, straight-forward calculation of standard deviation was prone
to run-off error, which could lead to negative variances and imaginary
standard deviation.  This new formulation uses the Welford algorithm which
is more resistent to loss of precision.  Note, however, that the algorithm
is only used when processing multiple ensemble members; if a single
deterministic run is processed, standard deviation is just set to missing.

### DIFF
--- a/lvt/core/LVT_DataStreamsMod.F90
+++ b/lvt/core/LVT_DataStreamsMod.F90
@@ -410,6 +410,10 @@ contains
     character*10                         :: stepType_mean
     character*10                         :: stepType_ssdev
 
+    ! EMK...For Welford algorithm
+    integer :: count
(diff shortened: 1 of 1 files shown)


emkemp:
DO NOT APPROVE THIS PULL REQUEST JUST YET.

Some discussion is probably warranted on whether the "smoothing exception list" should have been removed.  If we decide to keep it, I can update this request to reinstate the list.

emkemp:
I believe this pull request is ready.  Further changes to what variables should be smoothed will be handled in a different pull request, and will be handled by scripts.

PR #449, opened 2019-12-28
Fixes the bug in LVT UA reader

### LINKED ISSUE #448: UA snow reader in LVT fails when UA files do not exist
Some of the deallocate statements are in the wrong location. The code fails with the following error: 
forrtl: severe (153): allocatable array or pointer is not allocated
Image              PC                Routine            Line        Source
LVT                0000000000C9C2D0  Unknown               Unknown  Unknown
LVT                0000000000B5025D  readuasnowobs_            153  readUASNOWObs.F90

sujayvkumar:
Moves the deallocate statements to avoid the code crash when
UA files that are being read do not exist.

Fixes #448


sujayvkumar:
The fix is very minor and I am accepting the pull request.

PR #450, opened 2020-01-10
Revise list of LIS variables to smooth

emkemp:
This changes a number of scripts for automating LVT in "557 post" mode.  New philosophy is to smooth a variable if:
(1) it is perturbed;
(2) it is derived from a perturbed variable; or
(3) it is output by an LSM, and thus may be affected by perturbations.

In addition, the run_ncks.py script is updated to support Tair_f_min for 3-hr JULES output.


jvgeiger:
It seems that the scripts were modified for running on discover instead of on conrad; in particular, the values for _NCKS_PATH and output.  Was this intended?

emkemp:
Yes, the change to _NCKS_PATH is intended.  As long as the Conrad path remains in the script as a comment, it is trivial to run the script on either Conrad or Discover.

PR #453, opened 2020-01-27
Jules5

shugongwang:
Hello Jim and Eric, 

I have finished the JULES containers for 5.0, 5.1, 5.2, 5.3, 5.4 and 5.5 (5.x) and corresponding tests. All of them worked successfully with my NLDAS domain test cases. Could you please review the code for merging it to the main repository? My test case directory is /discover/nobackup/swang9/Models/JULES/TEST/.  

Thanks, 

Shugong


shugongwang:
Hopefully, I didn't break anything :)

jvgeiger:
The jules.5.x does not have the updated couple_jules_to_lis.sh script.  And it has some of the old files that we remove because of the new script.

jvgeiger:
Also, jules.5.0 does not compile.  You must rename ldtsi to usafsi in all files, subroutines, and variable names.  Please see surfacemodels/land/noahmp.4.0.1/da_usafsi for an example.

jvgeiger:
Your TEST directory contains OUTPUT and OUTPUT.good directories.  The results differ but not by much.  Which one should I compare against?

Also your lis.config files run from 1979 through 2018; although you seem to abort the runs within the first year.  I am going to change your lis.config files to run for one month.

shugongwang:
Hi Jim,

I think that my git version should be “too” old. Even in my Noah-MP directory, the variables and subroutines are named with ldtsi. My question is whether the renaming can happen on your side? Otherwise, if I update my local repository, it is possible that I lose my new code? I am a little bit scared.

Please advise,

Thanks,

Shugong

shugongwang:
Hi Jim,


The outputs in OUTPUT.good was produced by LIS WITHOUT “-fp-model strict” flag. The outputs in OUTPUT are produced with the flag. The differences are caused by the flag.



My jobs just run in 1 hour in the debug queue. This is why the jobs were aborted.



Thanks,



Shugong

jvgeiger:
I do not understand your comment regarding ldtsi.  I am looking at your branch, and I see the ldtsi to usafsi updates.

shugongwang:
Ah? I am confused too. I will check again!
Thanks,

jvgeiger:
The couple_jules_to_lis.sh scripts do not have my updates for compiling without MPI.

jvgeiger:
I am able to run these six versions of JULES and reproduce your results.  I made several updates to this branch, and I am still able to reproduce your results.

But for JULES 5.2, 5.3, and 5.4, the results are questionable.  For example, I see qle=0 for the whole domain for 2 months.

I also want to confirm the LAI is supposed to be a constant 5 for the whole domain.

I also compiled/ran JULES using the GNU compilers (which is necessary for the upcoming tutorial).  To me, some of the scatter plots in the JULES 5.x results are too spread out for changing compilers.

Please use this modulefile for any GNU testing: /home/jvgeiger/privatemodules/lisf_7_gnu_4_9_2

Please update your jules5 branch before committing any new updates.

plot_qle_tavg 01 swang9

fenxi_gpp_tavg 59 gnu_vs_intel

shugongwang:
Hi Jim,

I know what is going on for “But for JULES 5.2, 5.3, and 5.4, the results are questionable. For example, I see qle=0 for the whole domain for 2 months.” There is a flag that we have to set.  I will do it quickly.

Thanks,

Shugong

Issue #454, opened 2020-02-04
Add new LDT user-derived irrigation map reader


karsenau:
Implement new reader in LDT that can support user-derived irrigation maps (more general reader).

emkemp:
This was fixed by pull request #455.  I am closing the issue.

PR #455, opened 2020-02-05
Feature/new irrigreader

karsenau:
This pull request covers a new LDT reader that supports user-derived irrigation fraction maps, and ticket issue #454 

New testcase that supports this new reader can be found here:
/discover/nobackup/projects/lis/Projects/LDT/LISF_Testcases/Param_IrrigUserDerived


karsenau:
Side note: User can now request a reviewer ...

emkemp:
I started working with this pull request.  Intel and GNU compilers detect no problems.  Still need to run the test case.

emkemp:
I added a sanity check to detect cases where the irrigation fraction map projection is not latlon.

Otherwise, code looks fine, and runs w/o issue with both Intel and GNU compilers.  I approve.

Issue #407, opened 2019-11-12
Small snow amounts not cleared in Noah3.6


dpsarmie:
Issue involves very small amounts of snow not being properly removed in Noah 3.6, which results in very small amounts of snow accumulation in sub-tropical and other regions. 

Since this fix is not a part of the official Noah 3.6 release, a config option in LIS will be added to allow the user to use a fix that was developed internally or to use the official Noah 3.6 release as-is. 
This issue does not seem to be present in other versions of Noah, so this option will only be applicable to version 3.6.

PR #459, opened 2020-02-13
Add residual snow fix option to Noah.3.6

dpsarmie:
This commit adds a fix to Noah.3.6 that will remove small amounts
of snow that should normally be cleared but are not in Noah v3.6.

This fix will zero out snow on a model point if it is below a certain
threshold. Since this fix is not an official NCAR release, the fix
is added as a config option.
The user must specify 'Noah.3.6 removal of residual snow fix: 1'
to turn the fix on. The fix is turned off (option 0) by default.

Resolves: #407

### DIFF
--- a/lis/configs/lis.config.adoc
+++ b/lis/configs/lis.config.adoc
@@ -7283,6 +7283,11 @@ height in meters of air temperature and specific humidity observations.
 `Noah.3.6 reference height for forcing u and v:` specifies the
 height in meters of u and v wind forcings.
 
+`Noah.3.6 removal of residual snow fix:` specifies whether or not
+a snow fix is used (1 - fix is active, 0 - inactive). The land-surface
+model was found to keep very small amounts of snow in rare cases. This
+fix zeros out snow values that are under a predetermined threshold.
+
 .Example _lis.config_ entry
 ....
 Noah.3.6 model timestep:                  15mn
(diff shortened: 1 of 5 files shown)


emkemp:
Hi Daniel:

Do you have a simple test case to demonstrate this new feature?

emkemp:
Hi Daniel:

I managed to test the modified code with a case I already had using Noah 3.6.

The existing code in noah36_readcrd.F90 is not quite ready yet.  I see three bugs:

(1) There is no check on whether the new "Noah.3.6 removal of residual snow fix:" is in the lis.config file.  The rc code from ESMF_ConfigFindLabel should be checked and handled (if the entry is missing, the snowfix values should be explicitly set to zero without calling ESMF_ConfigGetAttribute).

(2) The [INFO] messages do not repeat the exact "Noah.3.6 removal of residual snow fix:" entry spelling.  This could cause confusion to the user if the setting is incorrect (e.g., if they want to zero out the snow, but are using an old lis.config file that doesn't have the entry).

(3) Since multiple snowfix settings exist (one for each grid nest), the [INFO] statements should include information on what value is used for which nest.

So far I only tested with Intel.  I will test with GNU compilers shortly.

emkemp:
Hi Daniel:

I rewrote the code in noah36_readcrd.F90 to require the new config option to be in the lis.config file, as well as to handle multiple nests.

I backed off on my objection to the wording of the general status statements, since your approach matches that taken elsewhere by LIS.  On reflection, the exact wording of the lis.config entry is only really required when an error occurs, and the modified code addresses that.

Otherwise, the code works for both Intel and GNU compilers.  I approve.

PR #422, opened 2019-12-02
Fix/410 ldtbuffer

karsenau:
Add new buffer area to LIS target domains in LIS when subsetting parameters and reinterpolating them to the new target grid. 

This new code was applied to remove a bug involved with curvilinear projections (e.g., Lambert conformal). 

Also, an outdated longitude ordering (from E->W) was discovered in some of the ipolates routines, which were corrected to allow longitudinal ordering to go from W->E, as LIS convention follows, fixing issues for projections, like latlon grids crossing the International Dateline.

Testcase was also setup and instructions provided.

Issue #410, opened 2019-11-13
LDT bug when processing coarse resolution data 


sujayvkumar:
When the LDT configuration involves the transformation of a coarse resolution data to a fine resolution target grid in curvilinear projection (lambert, for e.g.), the code seems to add artificial gaps along the boundary in the processed output. This is likely caused by the approach in LDT where the input data is first subset before interpolating and retrojecting to the fine resolution grid. The coarse resolution subset may not be sufficiently large to avoid causing gaps along the boundaries. When different interpolation options are on, different types of gaps are seen in the output.

karsenau:
I will work on this and test for both Lambert and latlon domains.

karsenau:
Update: 
Latest code modifications were successful in:

1)  adding a domain buffer to the target LIS grid for when subsetting parameters for Lambert projection - fixing the field gap issue;
2) fixed some issues encountered with latlon domains crossing the IDL; and
3) was able to exactly replicate a latlon CONUS domain with no changes to the output fields (at least for Noah parameters).

Will conduct a few more domain and projection tests prior to final push of the code updates.  Also, will prepare a testcase for these updated fixes.

karsenau:
Note: Latest commit for this issue #410 relate only to the LDT part of the code. No similar updates made at this time to LIS.

karsenau:
Quick update: Tested for an IDL-crossing domain with Lambert projection and all looks good.

karsenau:
This issue should now be resolved with the latest code commits to the master branch (accepted).

karsenau:
Add new buffer pixel options in the LDT config file to allow users to set the number of pixels in the x- and y-directions that add a buffer to the subsetted parameter target domain for transformations (e.g., from latlon to lambert grids).

karsenau:
Final code is now committed and pull request made.

karsenau:
New fix to LDT grid domains (Issue #410)

 This commit provides new updates to LDT in relation
  to the "buffer" outer domain option and support for domains
  that cross the International Dateline (IDL).

 - Resorted code to original lisdom_min|max calculations,
    which more correctly specifies outer domain required
    for curvilinear projections (e.g., Lambert);
 - Removed recently added code for these buffer and IDL domain
    issues, and replaced with simple code addition for when
    accounting for crossing the IDL.
 - Simplified the buffer code option and left the options in,
    however, they will not be triggered unless the user wants
    to turn them on. Default option is "off".
[...]

PR #460, opened 2020-02-18
Fix/410 ldtbuffer

karsenau:
Added new LDT config file options and documentation to support user-specified x- and y-direction number of buffer points when needing to add a buffer for grid transforms and such. This completes final commits for issue #410 and #422

### DIFF
--- a/ldt/configs/ldt.config.adoc
+++ b/ldt/configs/ldt.config.adoc
@@ -407,6 +407,28 @@ The default value is 1.
 Add buffer to parameter grid domain:     0
 ....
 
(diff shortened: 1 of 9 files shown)


karsenau:
An LDT testcase is provided here that provide an example of these new options:
/discover/nobackup/projects/lis/Projects/LDT/LISF_Testcases/Param_LambertCheck

ldt.config
ldtlog_prelis.0000
lis_input_d01.nc

emkemp:
Compilation and run-time tests show no issues.  Code looks fine, documentation looks fine.  I accept.

Issue #464, opened 2020-02-20
Change Noah-MP 3.6 irrigation rate to be constant over irrigation window


jerlingi:
The current irrigation code in Noah-MP 3.6 checks the moisture availability threshold every integration time step. The proposed change will return to a single check of the soil column at the beginning of the irrigation window and will result in a constant rate applied for the entire irrigation window.

emkemp:
This was fixed by pull request #469.  I am closing the issue.

PR #469, opened 2020-02-26
Adjust irrigation window timing in NoahMP3.6

jerlingi:
This pull request reverts noahmp36_getirrigationstates to checking the soil moisture availability in the soil column for sprinkler irrigation once at the beginning of the irrigation window and applying a constant rate over irrhrs. It fixes a previous bug where this check was performed but then the irrigation was turned off at subsequent integration time steps. 

This also adjusts the irrigation timing so that the irrigation falls neatly into LIS_HIST irrigation files for the irrigation window (see figure). 

This resolves issue #464 

Testcase in /discover/nobackup/projects/wldas/share/noahmp36_irrigation_bugfix/

SM_TestCase_y103_x27_Zoom_Shift


emkemp:
Made some minor code cleanups.  Otherwise, code looks acceptable.  Code compiles with both Intel and GNU compilers.  I approve.

Issue #466, opened 2020-02-22
Document "AGRMET GALWEM nominal resolution (km):"


jvgeiger:
We need to add documentation for "AGRMET GALWEM nominal resolution (km):".

emkemp:
This is addressed by pull request #474

emkemp:
Pull request #474 was merged into support/lisf-557ww-7.3.  I will leave this issue open until the changes migrate into master.

dmocko:
This change has been migrated into master, so I am going to close this issue.

PR #474, opened 2020-03-02
Added entry for GALWEM nominal resolution in kilometers.

emkemp:
Currently only 17 and 10 km are supported.


emkemp:
This addresses issue #466.

PR #475, opened 2020-03-04
Enable build tools to use either Python2 or Python3

jvgeiger:
The build tools for LISF (plugins.py and makedep.py) have been updated to work with either Python 2.7 or Python 3.

I plan to support both Python2 and Python3 until after the LISF public 7.3 release and after the migration to SLES12.  Then the build scripts will be updated to Python3 only.

### DIFF
--- a/ldt/make/makedep.py
+++ b/ldt/make/makedep.py
@@ -5,6 +5,26 @@ import os
 import re
 import glob
 
+#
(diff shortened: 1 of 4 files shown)


emkemp:
Code tested with both Python 2.7 and Python 3.6.  I have no objections.  Approved.

Issue #511, opened 2020-04-30
Incomplete documentation for the irrigation code


sujayvkumar:
There are several new irrigation options which are not included in the master lis.config 

Irrigation GVF parameter 1:
Irrigation GVF parameter 2:

These config options are not intuitive - Please come up with better config entries for these  (minimum threshold, maximum threshold etc.)

sujayvkumar:
The following log statement in noahmp36_getirrigationstates.F90 should be disabled (it  inundates the LIS log file when irrigation does not occur)

sujayvkumar:
! If we are outside of the irrigation window, set rate to 0
     if ((ltime.gt.shift_otimee).or.(ltime.lt.shift_otimes)) then
       irrigRate(t) = 0.0
       write(LIS_logunit,*) '[INFO] ',ltime, &
            'is outside the irrigation window'     
     endif

jerlingi:
Hi Sujay,

Sorry about the stray comment making it in to the final version! in SRC/lis/configs/lis.config.adoc, I had included a brief description of these parameters:

`Irrigation GVF parameter 1:` specifies growing season for
irrigation in Noah-MP (set to 0.40 for Noah 3.3 irrigation)

`Irrigation GVF parameter 2:` specifies growing season for
irrigation in Noah-MP (set to 0.00 for Noah 3.3 irrigation)

I agree that they need a better description, but should they be also somewhere else too?

karsenau:
I think we could provide more specifically what the two parameters represent and the source for this update to the original work implemented by Mutlu and Hiroko.

Here is some additional information that we could pull from and provide more background on these two parameters in the lis.config.adoc file.  

[...]

PR #515, opened 2020-05-05
Update irrigation documentation

jerlingi:
This pull request updates the Noah-MP 3.6 irrigation to comment out an unnecessary print statement. It also clarifies the GVF parameters in the lis.config.adoc file. 

These fixes address #511


bmcandr:
Hi Jessica,

Is there a good reason to keep the commented call to `write(LIS_logunit, *)` in `noahmp36_getirrigationstates.F90`? The preference would be to remove it if it's not necessary to avoid an accumulation of commented code. Let me know what you think!

-B

jerlingi:
Nope! Feel free to delete it!

Issue #524, opened 2020-05-28
Add support for the joint entropy metric in LVT


sujayvkumar:
https://en.wikipedia.org/wiki/Joint_entropy

PR #526, opened 2020-05-28
Implements the joint entropy method

sujayvkumar:
This commmit adds the support for the joint entropy metric
calculation

Related to #524


sujayvkumar:
A testcase can be found at : /discover/nobackup/svkumar/LISF_testing/524

bmcandr:
Hi @sujayvkumar 

The code looks clean and testing on Intel 18 and 19 as well as GNU4 produced identical results with no compile or runtime errors. Before I merge, I have a question: should Joint Entropy be added to `lvt/configs/METRICS.TBL`?

sujayvkumar:
The documentation has been updated as requested.

bmcandr:
Thanks @sujayvkumar. And thanks for also adding the metrics from recent PRs to the docs as well! I'm still getting a sense of which variables should be explicitly described in the documentation.

This PR looks to be in good shape and I am therefore merging.

PR #531, opened 2020-06-09
Add 1 km MERIT DEM (upscaled) for topography parameters 

yyoon4:
**MERIT Hydro DEM***

Update the DEM parameters for consistency with new HyMAP2 parameters.
: Firstly, 1-km global DEM was reproduced based on the 90-m MERIT DEM. MERIT DEM is only available above the 60S. Below the 60S, 1-km GTOPO30 is used. 

*Dataset available from http://hydro.iis.u-tokyo.ac.jp/~yamadai/MERIT_DEM 

**Example for LDT configuration**
#Topography parameter inputs
Elevation data source:                  "MERIT_1K"
Elevation number of bands:              1
Elevation map:                               ./MERIT
Elevation fill option:                      none
Elevation fill radius:                           5
Elevation fill value:                            0

Slope data source:                        "MERIT_1K"
Slope number of bands:                  1
Slope map:                                    ./MERIT
Slope fill option:                           none
Slope fill radius:                              5
Slope fill value:                               0

Aspect data source:                     "MERIT_1K"
Aspect number of bands:                 1
Aspect map:                                 ./MERIT
Aspect fill option:                        none
Aspect fill radius:                           5
Aspect fill value:                         3.14159

Topography spatial transform:           average
Topography map projection:              latlon
  
**Testcase:**
/discover/nobackup/yyoon4/lis7/Testcase/ldt_merit_dem

PR #532, opened 2020-06-10
LDT: HyMAP2, Support flexible grid setting (dx~=dy)

yyoon4:
Updated LDT hymap2 codes to support a flexible grid setting (especially to support current 557WW 10-km LIS run), as well as fixing minor typos.

**Testcase**:
/discover/nobackup/yyoon4/lis7/Testcase/ldt_hymap2

### DIFF
--- a/ldt/params/HYMAP/HYMAP_parmsMod.F90
+++ b/ldt/params/HYMAP/HYMAP_parmsMod.F90
@@ -16,6 +16,7 @@ module HYMAP_parmsMod
 !  29 Mar 2013: Sujay Kumar; Initial implementation
 !   2 Dec 2015: Augusto Getirana: Included drainage area and basin maps
 !   1 Nov 2017: Augusto Getirana: Included flow type maps, baseflow and surface runoff dwi maps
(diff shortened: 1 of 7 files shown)


yyoon4:
Hi @bmcandr 

1. File ldt/params/HYMAP/read_HYMAP_fld_height.F90:
Yes, you're correct!!

2. File ldt/params/HYMAP/read_HYMAP_grid_elev.F90:
Lines 65-74: remove #if 0 block if not needed -> **keep it for later**

Thanks,
Yeosang

bmcandr:
Thanks @yyoon4 for confirming my thinking re: the typo fix!

The code looks good and I have successfully reproduced output with Intel *and* GNU compilers. For the GNU test run, I followed your lead from #531 and modified `ldt.config` to use the NCEP-native dataset instead of Barlage's for max snow albedo. **Before I merge please confirm that this was an acceptable modification to your test case.**

yyoon4:
Hi @bmcandr 
Thanks for your review! Yes, you can use the "NCEP-native dataset" for max snow albedo!!

Thanks again,
Yeosang

bmcandr:
Thanks @yyon4. I'll go ahead and merge.

Issue #534, opened 2020-06-12
Bug in LIS soil moisture DA QC for Noah-MP-4.0.1


dmocko:
There is a bug in the LIS soil moisture DA quality control (QC) for the Noah-MP-4.0.1.

Good news, though, we have a fix ready to go, and this bug does not affect Noah-MP-3.6 or any version of Noah-3.X.

PR #535, opened 2020-06-12
Fix LIS soil moisture DA QC checks for Noah-MP-4.0.1 LSM

dmocko:
This commit fixes a bug in the soil moisture DA quality
control (QC) codes for the Noah-MP-4.0.1 LSM.  The code
was setting the value for all patches of variables SMCMAX
and SMCWLT to the same value.  This fix correctly sets
the values for all patches for these variables.

A review was done for Noah-MP-3.6 and Noah-3.X quality
control soil moisture DA codes, and it was determined
that no fixes are required to these other LSMs.

Resolves: #534


dmocko:
After merging into the support branch, this fix should also be merged into the master branch.

bmcandr:
Hi @dmocko,

Looks like a nice simple fix! It sounds like you've already tested these changes; is there a test I should run to verify before merging?

dmocko:
@yhkwon81 ran a test, and the 557WW LIS group reviewed it.  My opinion is that you also don't need to run a test before merging.

bmcandr:
Works for me!

dmocko:
I realize I also asked @jvgeiger for a review, but if he doesn't have time for it now, I would like to request that we can go ahead and merge this pull request now anyway.

Issue #525, opened 2020-05-28
Implement the relative entropy metric in LVT


sujayvkumar:
This metric is also called the Kullback-Leibler divergence 
https://en.wikipedia.org/wiki/Kullback–Leibler_divergence

PR #537, opened 2020-06-16
Implements the relative entropy metric in LVT

sujayvkumar:
This commit includes the implementation for the Relative Entropy or
Kullback-Leibler divergence metric. This metric is a measure of
how one probability distribution is different from another probability
distribution

Relative Entropy values are calculated as:

     H(X)= Sum_x  P(x)log [P(x)/Q(x)]

related to #525


sujayvkumar:
A testcase can be found at /discover/nobackup/svkumar/LISF_testing/525

bmcandr:
Thanks, everything looks to be in in order so I'll go ahead and merge.

PR #539, opened 2020-06-18
Fix/noahmp401 soilmda

yhkwon81:


jvgeiger:
Thanks Brendan. Yes, that comment is the justification for the update.

PR #542, opened 2020-06-28
Updated LVT to have more flexible time lag computations

### LINKED ISSUE #541: Improve the flexibility of time lag computations in LVT
The time lag computations are currently only supported for the second data stream. Make this more flexible so that additional datastreams can be run at a user specified time lag

sujayvkumar:
The time lag specification has been extended for all three
datastreams

Resolves #541

### DIFF
--- a/lvt/metrics/LVT_MinTimeMod.F90
+++ b/lvt/metrics/LVT_MinTimeMod.F90
@@ -293,19 +293,21 @@ contains
     type(LVT_statsEntry)    :: stats
     type(LVT_metricEntry)   :: metric
 !EOP
-    integer    :: t,k,m,m_k,o_k,tind
+    integer    :: t,k,m,m_k,o_k,tind,kk
     integer    :: time
 
 !----------------------------------------------------------------
 !It is assumed temporal lag is not chosen in this setup
 !----------------------------------------------------------------
-    if(LVT_rc%tlag(1).gt.0) then 
-       write(LVT_logunit,*) "[ERR] "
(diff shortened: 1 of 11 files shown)


sujayvkumar:
A testcase is at :
/discover/nobackup/svkumar/LISF_testing/541

sujayvkumar:
@bmcandr : To answer 3), Yes, the changes are needed to allow three data streams. Most LVT analyses only require two data streams, but there are some metrics that require three data streams. e.g. triple collocation metric.

lvt/metrics/LVT_MinTimeMod.F90
```diff
@@ -293,19 +293,21 @@ subroutine diagnoseSingleMintime(model, obs, stats,metric)
...
 !----------------------------------------------------------------
-    if(LVT_rc%tlag.gt.0) then 
-       write(LVT_logunit,*) "[ERR] "
-       write(LVT_logunit,*) "[ERR] Non-zero temporal lag specification is not "
-       write(LVT_logunit,*) "[ERR] supported for MinTime metric"
-       call LVT_endrun()
-    endif
-
+    do k=1,LVT_rc%nDataStreams
+       if(LVT_rc%tlag(1).gt.0) then
```
bmcandr, line 303: `kk` is initialized, but the value check in the if statement needs to be updated to use it (see `LVT_MaxTimeMod.f90`).

bmcandr:
I was able to reproduce the target output with the latest commits so I will go ahead and merge.

PR #546, opened 2020-07-06
Fix bug in index of loop in LVT_MaxTimeMod.F90

dmocko:
The do loop is indexed by "k", but the new variable "kk" is used within the loop.  The corresponding file LVT_MinTimeMod.F90 correctly has this do loop indexed by "kk".

This bug was introduced by pull request #542 .


bmcandr:
Thanks for catching this @dmocko. The test case ran fine for me; did an error occur on your end?

Regardless, I'll go ahead and merge.

dmocko:
No, I did not encounter an error.  Have not run LVT today.  I just noticed the difference when updating my code.

bmcandr:
Gotcha. Good eye.

PR #548, opened 2020-07-08
SMAP QC flag was removed

yhkwon81:
LIS does not ask "SMAP QC flag" for SMAP_NRTsm and
always turns on SMAP QC


dmocko:
I'm fine with this change, and think it can be merged.

PR #568, opened 2020-08-07
Fix/jules 1snow

emkemp:
Added separate namelists for running JULES with single snow layer instead of 3 snow layers.

The single layer is required for warm-starting LIS-JULES with UM Dump files in 2017 and 2018 when UM was still using a single snow layer.  The 3-layer configuration (PS41) is recommended for Air Force Ops.


emkemp:
Okay, I believe this is ready for review and (hopefully) acceptance.

shugongwang:
Thanks Eric! I have reviewed the namelist files and all look right.

Issue #180, opened 2019-06-26
Remove hard-coded options in IMERG LVT reader 


dpsarmie:
The current IMERG reader in LVT is hard-coded to work with only version V05B and the half-hourly product of IMERG data.
Two new config options will be added to allow the user to specify the version number of IMERG (similar to #146) and the temporal frequency of the IMERG data that is being used (either half-hourly, daily, or monthly).

dpsarmie:
Update to this issue:
I will fix the hard-coded filename issue in LVT, but scrapping the daily/monthly part of the ticket. 
Adding a daily or monthly option would require a new reader since the daily and monthly files are different file formats. For now, you can take the 30-min data and compile daily and monthly files as needed.

If there is a big need for a IMERG daily/monthly reader, then it can be revisited later.

PR #571, opened 2020-08-24
Fix LVT IMERG hard-coded filename option

dpsarmie:
The IMERG reader in LVT was still hard-coded to V05B and an old
filename extension convention. This fix will add an option for the
user to specify the version and product of IMERG data (similar to
the IMERG reader in LIS). The default options will be set to version
V06B and 'final' product.

Resolves: #180

### DIFF
--- a/lvt/configs/lvt.config.adoc
+++ b/lvt/configs/lvt.config.adoc
@@ -1246,12 +1246,15 @@ CPC PCP use real time data:               0
 
 `IMERG data directory:` specifies the location of the GPM IMERG precipitation product
 
-`IMERG version:` specifies the version of the GPM IMERG precipitation product. The valid options are: early, late, or final
+`IMERG version:` specifies the version of the GPM IMERG precipitation product. Most current version is V06B.
+
+`IMERG product:` specifies the GPM IMERG precipitation product. Current options are: early, late, or final.
 
 .Example _lvt.config_ entry
 ....
 IMERG data directory:            ../IMERG
-IMERG version:                   final
+IMERG version:                   V06B
+IMERG product:                   final
 ....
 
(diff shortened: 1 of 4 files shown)


dpsarmie:
Ok, fixes should be in the updated files.

dpsarmie:
Hey @bmcandr, I'm starting to put together a testcase for the IMERG reader (/discover/nobackup/dpsarmie/work/gitLIS/LVT_IMERG_Fix/IMERG_Testcase) but Grads isn't installed on SLES12. 
I wanted to check in to see if we're still making Grads files for testcases? I can log onto the SLES11 system and make them but wanted to check in before doing so.
Thanks.

bmcandr:
Hi @dpsarmie the Grads files would be helpful to verify the test run output looks good. Thanks for putting this testcase together!

dmocko:
Hi @dpsarmie - You can run GrADS on both SLES11 and SLES12.  You don't load a module.  Just add this to your PATH, and everything else should work:

/discover/nobackup/projects/gmao/share/dasilva/opengrads/Contents

David

bmcandr:
Thanks @dmocko! I had been switching over to SLES11 to run Grads as well so this is very helpful.

dmocko:
I often will switch to SLES11 to run GrADS because X11 can be slow on SLES12.  Plus the SLES11 login nodes aren't used as much now, so the load is usually much lower than the SLES12 login nodes.

dpsarmie:
Thanks @dmocko for the tip!

@bmcandr Testcase is here: /discover/nobackup/dpsarmie/work/gitLIS/LVT_IMERG_Fix/IMERG_Testcase
Let me know if you need anything else or run into any issues.

PR #575, opened 2020-08-25
update the testcase creation guide

bmcandr:
* Cleans up formatting
* Adds additional guidance

I've updated this guide based on my experience cleaning up existing testcases for inclusion in the test suite. Feedback welcome!


jvgeiger:
Thanks.  This looks good.

I ran into an error when I tried to process this file into a PDF document.  I will look into it.

bmcandr:
Fixed the error! The column formatting on the file table wasn't set properly for PDF conversion, but showed up fine when viewed as HTML.

jvgeiger:
Thanks!

I prefer .ctl files, but .xdf files are acceptable.

First an .xdf file is a simpler way to write a .ctl file for NetCDF output.  An .xdf file will not work for our binary or GRIB formatted output.

I prefer .ctl files for NetCDF data because they are explicit and allow me to fully automate generating plots.  But if I have problems with an .xdf file, I can quickly create a .ctl version.

bmcandr:
Thanks for the explanation @jvgeiger. I went ahead and removed the references to .xdfs to simplify things.

If everything else looks alright, I'll merge tomorrow morning.

jvgeiger:
Everything looks good.

Do you prefer admonitions like "Note" and "Important" to be spell out with words or with symbols (i, !)?

bmcandr:
I think I prefer the symbols. They take up less space and the color makes the admonitions stand out a bit.

Should I add the `:icons: font` attribute to enable them?

jvgeiger:
Yes.  Sounds good.

PR #577, opened 2020-08-27
Fix HYMAP2 routing write ensemble restart routine to compile serial

dmocko:
Small changes to the routine that writes the HYMAP2 routing scheme
ensemble restart files that allow LIS to compile both for parallel
and serial modes.  Previously, LIS would not compile in serial mode.

Two changes were made:

1) Move the "#if (defined SPMD)" compile directive outside a do loop.

2) Include the second ensemble dimension of the "var" variable.

Thanks to @nmizukami for finding this issue, identifying a fix, and reporting it.


bmcandr:
I successfully compiled serially so I'll go ahead and merge. Thanks @nmizukami!

PR #581, opened 2020-09-04
Fix jules50_qc_soilmobs

jvgeiger:
The elseif block for "elseif (l_aggregate)" should be moved to the end
of the whole if block.  Otherwise, QCs for sneqv_obs and fsno_obs
do not work ever when l_aggregate = true.


jvgeiger:
This was reviewed by Yonghwan.

PR #222, opened 2019-08-12
Feature/asyncsfout from PETTT project

emkemp:
Includes experimental asynchronous I/O code from PETTT project (Dan Rosen). This was fetched from NESII/LISF.git, and includes tag LISF_pettt_lis_0_5.

I don't know if this should go in support/lisf-557ww-7.3 or master.  For now, I pick master.

PR #584, opened 2020-09-08
Feature/async-io from PETTT project (revised)

bmcandr:
PR #222 contains commits unrelated to the asynchronous ~~I/O~~ *output* feature. I am opening this new PR based on commit 3149955 which contains only the changes relevant to this feature.

Based on testing, this feature appears incomplete and requires further discussion. Notes to follow.


bmcandr:
## Code Review
* In `lis/core/LIS_coreMod.F90`:
    * Lines 875-876: In the `#if (defined SPMD)` block `call MPI_FINALIZE(ierr)` is commented out in new code, but not in master. (Could this be related to issue where slurm job does not end after LIS run completes?)
    * If not needed, delete `#if ( defined SPMD )` block altogether?
* In `lis/core/LIS_cplMod.F90`:
    * Documentation needed:
        * Add description to module and subroutines.
            * Generally appears to couple ESMF and LIS to initialize, broadcast, and finalize use of ESMF route handles.
        * Describe `Write RH`/`Read RH` configuration options in `lis.config.adoc`?
            * See lines 102-111
    * Lines 257-258, 271-272: route handle file written to and read from hard-coded filename `LIS_CPL.RH` (if `Write RH: true` and/or `Read RH: true` options are selected in `lis.config`)
* In `lis/core/LIS_domainMod.F90`:
    * Should additions beginning at line 359 be placed in `if` block that only runs when async I/O is enabled?
* In `lis/core/LIS_histDataMod.F90`:
    * Line 5962: Add option 4 (irrigation) to documentation (see line 5978)
* In `lis/core/LIS_historyMod.F90`:
    * Incorrect stat summary output may have to do with this file? (see below for explanation)
* In `lis/offline/lis_esmf_comp.F90` (*SIGNIFICANT CHANGES MADE*):
    * Update docstrings to reflect filename. Looks like docstrings were copied over from `lisdrv.F90`.
* In `lis/offline/lisdrv.F90` (*SIGNIFICANT CHANGES MADE*):
    * Lines 66-67: remove commented code if not needed
    * Line 82: hard codes LIS config filename as `lis.config` -- using other filenames will result in failed run ⚠️
        * Additional hardcoded values exist throughout this file (e.g., lines 88, 119, 126, 191, 197)

## Notes about Running

* ESMF 8+ required for RouteHandle functionality (module avail @ `/discover/nobackup/bmcandr1/async-module/lisf_7_intel_19_1_0_166_ESMF8`)
* New options required in `lis.config` to enable/use async I/O. Place :

    ```sh
    Auxiliary components: SFOUTPUT  # SFOUTPUT must be in all caps (update code to make case insensitive?)
    LIS petlist bounds: 0 3         # Example values for a 2 x 2 processor scheme
    Aux petlist bounds: 4 7         # to add 4 processors for writing output (does not need to match LIS petlist dimensions and cannot overlap)
    Write RH: true                  # will run init, write route handle to LIS_CPL.RH (optional; default=false)
    Read RH: false                  # set true to read LIS_CPL.RH (optional, but LIS_CPL.RH must exist; default=false)
    ```
    Two ways to run:

    * Run once with *only* `Write RH: true` to generate ESMF routehandles and write to `LIS_CPL.RH`, then run with *only* `Read RH: true` to generate output.
    * Run with *both* `Write RH` and `Read RH` set to false to generate routehandles inline (*per Dan, no performance gain*).
    
    When running in parallel, arguments `--ntasks` and/or `-np` must equal the petlist upper bound + 1 as they are zero-indexed (e.g., 8 for the above example). This should be added to documentation.

## Testing Results
The results discussed below are the result of testing the new runmode with a few LSM testcases from the internal suite. I encountered some issues that need attention.

* ⚠️ LIS does not exit properly when running with `Read RH: true`. Interactive MPI runs must be ended manually and slurm jobs only end when they're revoked for exceeding job time limit. In some cases the run generates all output files contained in target, but in others only partial output is produced. For example, the `noahmp36_gwon` testcase will only produce SURFACEMODEL output files for the 3 days in the first month when it should produce 12 months of output. Further, output stops being produced after several minutes even if the job is allowed to continue for the allotted time. I can't find any errors that might help track down the source.
* For runs that appear to complete successfully (based on lislog) NetCDF output are ~identical to targets. The `*.d01.stats` files are not produced properly, however. These files contain empty statistical summaries for run timesteps. The stats summaries output then appears to restart, but the values do not match target. (This is not an issue of new output being appended to a leftover stats file as it appears in fresh output.)
    * Issue may lie somewhere in `lis/core/LIS_historyMod.F90` (section ending @ line 479?) and/or `lis/core/LIS_histDataMod.F90`?
* LIS log files contain null characters (e.g., `^@`) where output filepaths should be. This appears even when output is produced properly. For example:

```sh
[INFO] Writing surface model output to:
    ^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@
```
----
Internal test suite cases with modified config files available here: `/discover/nobackup/projects/lis/PR_TESTCASES/222/async-testcases` (to use the included LIS executable use the ESMF8 module linked above).

bmcandr:
This feature is being reworked to be more generic. If needed, the original work by Dan Rosen is needed it is available on the `feature/async-io` branch of my fork of LISF and can be cloned with:

```sh
git clone -b feature/async-io https://github.com/bmcandr/LISF
```

See my comment at the top of this PR thread for more context.

Issue #569, opened 2020-08-10
Create option to change max SWE depth in glacier model (NoahMP 4.0.1)


dpsarmie:
Currently, the glacier model that is part of Noah MP 4.0.1 has a hard-coded cap of 2000 mm for SWE. 

LIS users would gain flexibility when simulating areas with deeper snowpacks by creating a customization option in the config file (leaving 2000mm as default).

PR #586, opened 2020-09-16
Add max snow depth option for NoahMP 4.0.1 glacier model

dpsarmie:
This commit adds a config option that allows the user to set the
maximum snow depth (in mm) limit that is used in the Noah MP 4.0.1 
glacier model. This value was hard-coded to 2000. If the user does not
specify a value, the max snow depth will be set to the default value of 2000.
This change only affects NoahMP 4.0.1.

Resolves: #569

### DIFF
--- a/lis/configs/lis.config.adoc
+++ b/lis/configs/lis.config.adoc
@@ -8359,6 +8359,10 @@ Noah-MP-4.0.1 LSM initial water storage in the aquifer (mm).
 specifies the Noah-MP-4.0.1 LSM initial water in the aquifer
 and in the saturated soil (mm).
  
+'Noah-MP.4.0.1 snow depth glacier model option:' specifies the
+maximum snow depth that is used in the Noah-MP-4.0.1 glacier model (mm).
+The default is set to 2000 mm.
+
 .Example _lis.config_ entry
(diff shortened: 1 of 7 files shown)


dpsarmie:
Testcase is here: /discover/nobackup/dpsarmie/work/gitLIS/Noah401_SnowDepthOpt/testcase/

bmcandr:
Hi @dpsarmie . The code and documentation looks good. The target output you shared appears incomplete, however. The logs do not indicate that the run completed successfully. My test run also produced a restart file (`LIS_RST_NOAHMP401_201902010000.d01.nc`) not present in the target. The `LIS_HIST` file produced *is* identical.

Take a look at the output of my test run and let me know if it looks good. It is located here: `/discover/nobackup/projects/lis/PR_TESTCASES/586/testing/OUTPUT`

dpsarmie:
Hey @bmcandr . Sorry about that. I added the correct output to the testcase directory in my drive in case you need it but yes your output looks identical to my output. Let me know if you need anything else.

bmcandr:
Thanks @dpsarmie! Output looks good. I'll go ahead and merge.

PR #597, opened 2020-09-28
Update configure defaults for 557WW

emkemp:
IMPORTANT:  DO NOT MERGE THIS UNTIL PR 201 HAS BEEN BACKPORTED TO support/lisf-557ww-7.3!

Among the changes in this pull request is configuring LDT/LIS/LVT to compile in little-endian.  This will not work unless PR201 is backported first (LDT needs to use big-endian for certain inputs, and PR201 addresses this.)

### DIFF
--- a/ldt/arch/Config.pl
+++ b/ldt/arch/Config.pl
@@ -111,11 +111,11 @@ elsif($opt_lev == 3) {
    $sys_opt = "-O3 ";
    $sys_c_opt = "";
 }
-print "Assume little/big_endian data format (1-little, 2-big, default=1): ";
+print "Assume little/big_endian data format (1-little, 2-big, default=2): ";
 $use_endian=<stdin>;
 chomp($use_endian);
 if($use_endian eq "") {
-    $use_endian=1;
+    $use_endian=2; # For many LDT input datasets
 }
 
 
--- a/ldt/arch/Config.pl
+++ b/ldt/arch/Config.pl
@@ -111,11 +111,11 @@ elsif($opt_lev == 3) {
    $sys_opt = "-O3 ";
    $sys_c_opt = "";
 }
-print "Assume little/big_endian data format (1-little, 2-big, default=2): ";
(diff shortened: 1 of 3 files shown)


jvgeiger:
Do we wish to make little endian the default for the master branch as well?

emkemp:
I think we should, for simplicity.

dmocko:
I'm going to say not right now, as we know many of our datasets are still big_endian on Discover.  But since we know the 557ww chooses little_endian, it should be the default for that branch.

jvgeiger:
Sorry.  Does this fix an issue or is it simply for convenience?

emkemp:
I thought the point of PR201 was to hardwire endianness by input file, making the configure option irrelevant.  Is this not the case?

dmocko:
@jvgeiger - This pull request is simply for convenience when configuring/compiling in the support branch.

@emkemp - Yes PR201 does hardwire endianness by input file, but it only covered a subset of all codes that read binary datasets.  It does cover all datasets used by 557WW, but I haven't confirmed that all LDT/LIS/LVT binary dataset reads are now hard-coded.

emkemp:
@dmocko Okay, well, in that case the default endianness should not be changed.  The value will not matter for 557WW work.

dmocko:
@emkemp - Sorry, I left out crucial info.  PR201 does cover all LDT datasets used by 557WW.  We know that LIS has to be run in little_endian for AGRMET mode, so I think LDT/LIS/LVT should all have default little_endian for the support branch, for consistency.

emkemp:
Okay, I revised the endianness defaults.  LDT is big endian, LIS is little endian (for AGRMET), and LVT is big endian.

emkemp:
Okay, that sounds reasonable.  We can rehash this again with LIS 7.4.

dmocko:
Agreed - we're overthinking this.  Thanks for the reminder about the documents.

PR #601, opened 2020-10-02
Feature/rst transform

LIS-navari:
####This Pull request is on behalf of Sujay Kumar###
I have validated the feature with several test cases.  

A restart transform capability was added to the LDT. The New feature 
enables the ability to generate restart files for a sub-domain
from the restart file of a larger domain.

Test case 
Step1 : run LDT to generate the LIS parameter file.
/discover/nobackup/mnavari/my_projects/nrl/europe.45k/procLSM

Step2: 
Run LDT to generate restart file for a subdomain
/discover/nobackup/mnavari/my_projects/nrl/europe.45k/rstTransEx

Step3: 
Run LIS to make sure the generated restart file contains all fields and they are correctly subsetted from the original restart file.
/discover/nobackup/mnavari/my_projects/nrl/europe.45k/ol


bmcandr (changes requested):
Hi @LIS-navari,


## Code Review
* In `ldt/core/LDT_domainMod.F90`:
    * Line 230: remove comment if it has been addressed
* In `ldt/core/LDT_rstTransformProcMod.F90`:
    * Line 57: add brief description of subroutine
    * Line 85: if statement checks whether `LDT_rc%rstsource.eq."LSM"`, but does not check for other cases (e.g., `Routing`).
        * Should there be a defensive case? (i.e., print error message if anything but "LSM" is used?)
    * Line 118: only coarse->fine?
* In `ldt/core/LDT_rstTransformProcMod.F90`:
    * Lines 378-380, 396-398: remove leftover debugging code?
    * Line 429-430, 443-444: replace calls to `print*` with `write(LDT_logunit,*)` and calls to `stop` with `LDT_endrun()`
    * Line 564: should this comment be addressed/implemented?
* In `ldt/domains/lambert/readinput_lambert.F90`:
    * Lines 62-71: what does this "duplicate check" do?
        * Comment @ line 63 suggests this is an incomplete implementation?
* Documentation needed in LDT Users Guide. How should this new feature be described in documentation? (i.e., does it warrant a new section similar to section "7.11. Ensemble restart model options")

## Compile & Testing Results

I was able to compile successfully with GNU4, Intel 18, and Intel 19 compilers. I successfully reproduced output on Intel compilers, but was unable to test on GNU4 because the testcase relies on HDF-formatted *max snow* albedo data. **Would it be acceptable to use the NCEP Native max snow albedo for testing on GNU4?**

bmcandr:
Update: I modified the step 1 LDT config file to use the NCEP Native max snow albedo and the code ran fine so I think we're OK there. The output differed slightly from the target, but I understand that's to be expected.

PR #604, opened 2020-10-08
Add check for routing model selection in DA subroutines. [FIX]

### LINKED ISSUE #591: LIS seg faults due to unallocated variable (LIS_rc%Routing_DAinst_valid)
**Description:** LIS seg faults when attempting to run Step 6 of the public testcase with a LIS executable compiled from the latest master branch (Intel 19). No error is printed to the log and the bad stack trace w/o line numbers is produced:

```sh
Caught signal 11 (Segmentation fault: address not mapped to object at address 0x4)
==== backtrace ====
    0  /usr/lib64/libucs.so.0(+0x1935c) [0x2aaabc4a235c]
    1  /usr/lib64/libucs.so.0(+0x19512) [0x2aaabc4a2512]
    2  ./LIS() [0x6e2b84]
    3  ./LIS() [0x6f9cda]
    4  ./LIS() [0xb38e7f]
    5  ./LIS() [0x591898]
    6  ./LIS() [0x12abddb]
    7  ./LIS() [0xd28ab6]
    8  ./LIS() [0x413d12]
    9  /lib64/libc.so.6(__libc_start_main+0xf5) [0x2aaab45d7725]
   10  ./LIS() [0x413c29]
===================
```
**Cause:**
* LIS faults at [line 1290](https://github.com/NASA-LIS/LISF/blob/59fd37fd437ab6db172e39bfc4270ed13a3cd38c/lis/core/LIS_routingMod.F90#L1290) in `LIS_routingMod.F90` when `LIS_rc%Routing_DAinst_valid` is checked in an if-statement in the `LIS_routing_DAgetStateSpaceSize` subroutine (called on [line 773](https://github.com/NASA-LIS/LISF/blob/03d136c068484928128423575b2e2cb5ad9abec3/lis/core/LIS_surfaceModelMod.F90#L773) in `LIS_surfaceModelMod.F90`). 
* The fault occurs because `LIS_rc%Routing_DAinst_valid` is unallocated when it is checked. It appears that `LIS_rc%Routing_DAinst_valid` is only allocated in the `LIS_routing_init` subroutine if `LIS_rc%routingmodel` *does not* equal "none" (see [line124](https://github.com/NASA-LIS/LISF/blob/59fd37fd437ab6db172e39bfc4270ed13a3cd38c/lis/core/LIS_routingMod.F90#L124) in `LIS_routingMod.F90`).

**How to reproduce:** run Step 6 of the public testcase with a LIS executable compiled from the latest master branch (Intel 19). A staged testing setup including a debuggable executable (`LIS.latest.debug`) is available here: /discover/nobackup/projects/lis/PR_TESTCASES/share/routing-bugfix

bmcandr:
Fixes a bug where LIS would seg fault when a routing model
value of "none" is used. The if-statements added here prevent
the execution of code that should only run for routing model values
*other than* "none".

Resolves #591

Compiled and tested with Intel 19, Intel 18, and GNU4 compilers using Step 6 of the public testcase.

### DIFF
--- a/lis/core/LIS_dataAssimMod.F90
+++ b/lis/core/LIS_dataAssimMod.F90
@@ -445,9 +445,11 @@ contains
              endif
           enddo
           do i=1,LIS_rc%ndas
-             if(LIS_rc%Routing_DAinst_valid(i)) then
-                call routingdainit(trim(LIS_rc%routingmodel)//"+"//&
-                     trim(LIS_rc%daset(i))//char(0),i)
+             if(LIS_rc%routingmodel.ne."none") then
+                if(LIS_rc%Routing_DAinst_valid(i)) then
+                   call routingdainit(trim(LIS_rc%routingmodel)//"+"//&
+                        trim(LIS_rc%daset(i))//char(0),i)
+                endif
              endif
           enddo
        endif
--- a/lis/core/LIS_routingMod.F90
+++ b/lis/core/LIS_routingMod.F90
@@ -651,9 +651,11 @@ module LIS_routingMod
 
     integer                :: m
 
-    if(LIS_rc%Routing_DAinst_valid(k)) then
-       call routingdagetobspred(trim(LIS_rc%routingmodel)//"+"//&
-            trim(LIS_rc%daset(k))//char(0),n, k, Obs_pred)
+    if(LIS_rc%routingmodel.ne."none") then
+       if(LIS_rc%Routing_DAinst_valid(k)) then
+          call routingdagetobspred(trim(LIS_rc%routingmodel)//"+"//&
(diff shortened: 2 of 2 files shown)


bmcandr:
Hi @jvgeiger, after looking over the code and documentation again I believe it is expected that these routines do nothing when LIS_rc%routingmodel is "none". The LIS Users Guide [lists "none" as an acceptable value](https://github.com/NASA-LIS/LISF/blame/master/lis/configs/lis.config.adoc#L974) for the `"Routing model:"` option and the code within these routines should only run when a routing model is selected.

@sujayvkumar can you confirm? (see Jim's question [here](https://github.com/NASA-LIS/LISF/issues/591#issuecomment-705650977) for context)

PR #609, opened 2020-10-16
Fix AWRAL compile and run errors w/ GNU4 compiler

bmcandr:
This pull request fixes compile and runtime errors in the AWRAL600 code that are thrown on SLES11 with the GNU4 compiler.

**GNU compile errors fixed:**
* In `read_AWRAL.F90`: pad character variables with whitespace on initialization to fix a `Different CHARACTER lengths` error. These variables are always `trim`med before use.
* In `AWRAL600_readrst.F90`/`AWRAL600_writerst.F90`: convert logical operators in preprocessor directives from `.OR.` to `||` to fix an `error: token "." is not valid in preprocessor expressions` error. This solution is consistent with preprocessor directives elsewhere in LIS.

**Runtime errors fixed:**
When using a GNU4-based executable, LIS would seg fault with an "invalid memory reference" error in the finalize routine when attempting to deallocate uninitalized pointers.
* In `AWRAL_lsmMod.F90` and `AWRAL_module.F90`: initializations changed from pointers to allocatables
* `if (ASSOCIATED...` checks changed to `if (allocated...` checks in the finalize routine

While fixing the above I noticed that in `read_AWRAL.F90`, a grid check incorrectly compared `AWRAL_struc%ncol` to `LIS_rc%gnr`. The check would therefore evaluate to false and end the run when the forcing grid was the same as the model (error only thrown with GNU4-based executable?). I also had to convert the logical operator ".neqv" to ".ne" as the former requires both operands to be of type LOGICAL while the operands here are INTEGER(4). (This fixes a `Operands of logical operator '.neqv.' at (1) are INTEGER(4)/INTEGER(4)` compiler error.)

These fixes have been tested with GNU4, Intel18, and Intel19 compilers. ✔️ 

**Testcase files are staged here:** /discover/nobackup/projects/lis/PR_TESTCASES/370/awral600_testcase

### DIFF
--- a/lis/metforcing/AWRAL/read_AWRAL.F90
+++ b/lis/metforcing/AWRAL/read_AWRAL.F90
@@ -114,7 +114,7 @@ subroutine read_AWRAL( order, n, findex, year, doy, ferror_AWRAL )
   timestep = doy
 
 !-- Check forcing is on the same grid as the model --!
-  if((AWRAL_struc(n)%ncol .neqv. LIS_rc%gnr(n)) .or. (AWRAL_struc(n)%ncol .neqv. LIS_rc%gnc(n))) then
+  if((AWRAL_struc(n)%nrow .ne. LIS_rc%gnr(n)) .or. (AWRAL_struc(n)%ncol .ne. LIS_rc%gnc(n))) then
      if(LIS_masterproc) then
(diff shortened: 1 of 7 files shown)


dmocko:
Thanks for finding and fixing these issues @bmcandr - and for testing them on the different compilers.

I approve the code changes, and will merge.

PR #612, opened 2020-10-21
Add details to how met forcing overlays work to User Guides

dmocko:
This commit adds additional details and explanation with an example
to how met forcing overlays work within both LDT and LIS to their
respective User Guides.


bmcandr (approved):
Thanks @dmocko! Looks good. I'll go ahead and merge.

PR #625, opened 2020-11-06
Add GALWEM support to USAFSI

emkemp:
Adds support for GALWEM 17-km or 10-km GRIB2 data as source of 2-m air temperature in USAFSI.

Includes updates to ldt.config documentation.


jvgeiger:
Hello Eric,

I have a few questions and comments.

In ldt/USAFSI/USAFSI_galwemMod.F90:

1. subroutine USAFSI_get_galwem_t2m

```
130           write(LDT_logunit,*) '[WARN] in grib_count_in_file: ' // &
131                'dataTime in USAFSI_get_galwem_t2m'
```

This warning message is incorrect.

2. subroutine USAFSI_get_galwem_t2m

```
384        if (found) exit
```

I believe that this line is should be `if ( k > nvars ) exit`.  The `while` loop will end when `found == .true.`.

3. subroutine USAFSI_get_galwem_t2m

```
389     if (.not. found) then
```

Does not make sense unless we change line 384.  As written, the code will not reach line 389 unless `found == .true.`.  At which point, the code will not execute the body of this `if` block.

4. subroutine USAFSI_get_galwem_t2m

```
402     return
```

This return statement is not needed.

5. subroutine interp_galwem_t2m

Can we compute the weights and neighbours just once, say in subroutine set_galwem_griddesc?

jvgeiger:
Regarding 5.  I see that this is being done just once.  This routine processes only 2m temperature.

emkemp:
Hi Jim:

Regarding 2:  The logic is designed to gracefully handle a corrupt GRIB file.  If after looping through a particular file and not finding 2-m temperature, the subroutine will roll back to an earlier forecast run.  This "roll back" option is part of the outer while loop, while the inner do loop (k = 1, nvars) handles a particular GRIB file.  The "found" logical flag should only be set to .true. if 2-m temperature is indeed found; and at that point, both outer and inner loops should be exited.

Bottom line:  I believe the existing logic is correct.  If 2-m temperature is not found (or not found with the right map projection, time, cannot be read, etc), the inner loop will continue cycling until all nvars are searched.  Found will still be false, so the outer loop will cycle.

emkemp:
Hi Jim:  

Regarding 3: Again, it is possible 2-m temperature won't be found after searching all the GRIB files.  In this case, the logic will cause the subroutine to return early, cleaning up allocated memory before it leaves.

emkemp:
Hi Jim:

Regarding 5:  Those interpolation weights are only need to be used once, so I don't think there's a need to move the code.

jvgeiger:
Hello Eric,

Sorry.  I did not take the roll back into account.  I now see that line 97 will abort the `while` loop in case everything goes wrong.  And now I understand the `if` at line 389.  Thanks for the explanation.

I am still confused by the `if` on line 384.  I believe that it is not needed.  The code leaves the `do k = 1, nvars` loop when the GRIB file cannot be opened, when `k > nvars`, or when 2 meter temperature is found.  All three cases are handled by the `enddo` at line 388.  For the first two cases, the code goes back to the top of the `while` loop and rolls back.  For the last case, the code found the temperature field.

What else did I not take into account?

And I agree with you regarding 5.

emkemp:
I removed the if statement in 3) above.

PR #628, opened 2020-11-10
Add lisf_7_intel_19_1_3_304 modulefile

jvgeiger:
This modulefile is for compiling and running LISF on discover.

This modulefile loads comp/intel/19.1.3.304, which resolves the issue
of getting segmentation fault errors from ifort/icc on discover.

### DIFF
--- a/env/discover/lisf_7_intel_19_1_3_304
+++ b/env/discover/lisf_7_intel_19_1_3_304
@@ -1,0 +1,168 @@
+#%Module1.0###################################################################
+
+proc ModulesHelp { } {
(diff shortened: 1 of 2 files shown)


dmocko:
Thanks so much for tracking this down, @jvgeiger 

Do we expect any issue with many of the libraries being compiled with intel-19.1.0.166, while using the intel/19.1.3.304 module?

jvgeiger:
All of my tests passed with the  intel/19.1.3.304 compiler and the intel/19.1.0.166 libraries.

bmcandr:
Glad to see a solution has been found! Thanks for creating these new modulefiles.

I will go ahead and merge.

PR #629, opened 2020-11-10
Fix metrics config entries in the LVT Users' Guide

dmocko:
This pull request fixes typos for the following three config
entries in the LVT Users' Guide:

Metrics computation frequency:
Metrics output directory:
Metrics output frequency:

Previously, these config entries were listed as "Metric "
(without the "s").

The LVT code requires these entries to be spelled "Metrics",
so have fixed the Users' Guide to reflect this.


bmcandr:
Thanks for making these corrections @dmocko! Looks good, I will merge.

PR #633, opened 2020-11-13
Make JULES snow depth and SWE available to AGRMET Ops code

emkemp:
The AGRMET code retrieves snow depth and SWE stored in LIS_snow_struc (defined in lis/core/LIS_snowMod.F90).  However, this data was not actually copied into the structure by the JULES 5.0 interface (unlike Noah 3.9).  This patch set fixes that oversight.

LIS-JULES functional tests have been performed in AGRMET Ops mode w/o DA, w/ SNODEP DA, and w/ USAFSI DA.  Log output showed the snowDepthQC test in the precipitation analysis was seeing positive snow depth.  An additional hack (not included in this patch set) was to copy the LIS_snow_struc snow depth into the precipitation analysis array for output.  Visualization showed the snow depth values to be in the correct geographic locations, and to show expected variations between the no DA, SNODEP DA, and USAFSI DA cases.

Simulations were also performed with the code compiled with strict checks (-2).  No runtime errors were detected.

TODO:  Similar changes for NoahMP 4.0.1.  That will be a separate pull request.

### DIFF
--- a/lis/surfacemodels/land/jules.5.0/da_snodep/jules50_setsnodepvars.F90
+++ b/lis/surfacemodels/land/jules.5.0/da_snodep/jules50_setsnodepvars.F90
@@ -76,42 +76,22 @@ subroutine jules50_setsnodepvars(n, LSM_State)
      LIS_snow_struc(n)%snowdepth = 0.0 ! Mean snow depth by grid id
      LIS_snow_struc(n)%sneqv = 0.0     ! SWE by tile
 
-     ! Logic below follows jules50_main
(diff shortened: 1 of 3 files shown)


dmocko:
Looks good to me.  Why do all three files include the new commented-out lines showing the logic?  Is this needed?

emkemp:
Good point.  I removed the dead code.

dmocko:
Thanks.  I approve of these changes.

Do @jvgeiger or @shugongwang want to also take a quick look before this code is merged?

PR #642, opened 2020-12-04
Minor tweaks to Noah-MP config entries for LIS Users' Guide

dmocko:
This commit gives some minor tweaks (clarifications) for two config
entries for the Noah-MP LSM.  One is for the reference height of the
forcing; the other is for an optional config entry to set a maximum
SWE value for Noah-MP-4.0.1 glacier points.

See also: #586


bmcandr:
Hi @dmocko,

Looks like I hit merge a bit too swiftly. There's a single quote where a backtick should be on this change:

`Noah-MP.4.0.1 snow depth glacier model option:' <--

Probably easier to just push the fix rather than reverting the merge. Sorry about that!

dmocko:
I just pushed a fix - thanks for catching that!  I fixed the first backtick, but forgot the other one.

PR #644, opened 2020-12-04
Correct AsciiDoc formatting

jvgeiger:


lis/configs/lis.config.adoc
```diff
@@ -2644,26 +2644,26 @@ Acceptable values are:
...
 |====
 
-For 'AGRMET ops', dates before 2017-10-05 use the
+For "`AGRMET ops`", dates before 2017-10-05 use the
 smops_dYYYYMMDD_sHH0000_cness.gr2 naming convention;
 dates on or after 2017-10-05 use the NPR_SMOPS_CMAP_DYYYYMMDDHH.gr2 naming
 convention.
 
-NOTE:: When reading the daily data feed, the naming convention is 'LIS'.
+NOTE:: When reading the daily data feed, the naming convention is "`LIS`".
```
bmcandr, line 2652: Change `NOTE::` to `NOTE:` so the admonition formatting renders properly.

bmcandr (changes requested):
Hey @jvgeiger,

In addition to the note fix described in the inline comment, I see a typo at the end of the note on line 316 in _lvt.config_ that could be fixed. Seems like the `SoilMoist` example entry should be inside the note?

jvgeiger:
Got it.  Thanks.

PR #647, opened 2020-12-07
Added ability to read WWMCA GRIB1 files instead of binary

emkemp:
This feature was requested by 557WW due to issues with transitioning to System 11 (apparently only WWMCA GRIB1 files will be available on that platform, instead of the legacy flat binary files currently used by LIS).  When turned on, the new functionality will search for WWMCA GRIB1 files first before falling back on binary files, if necessary.  When turned off, the existing behavior of searching for binary only is preserved.

The new lis.config entry (now required) is:

AGRMET WWMCA GRIB1 read option: 1 # 1 for searching for GRIB1, 0 for just searching for binary

and this entry has been added to the sample files in lis/configs/557WW-7.3-FOC.

Note 1:  WWMCA GRIB1 files only exist in the AGRMET archive beginning 12Z 4 Jul 2020.

Note 2:  This should be added to LISF 7.3 before being migrated to the master branch.


emkemp:
POSTPONE THIS PULL REQUEST

I think I just found a logic error.  Will address.

emkemp:
I believe this is fixed now and ready for review.

Issue #409, opened 2019-11-12
LIS_writeModelOutput '[INFO] Writing surface model output to:  ' produces garbage


jvgeiger:
The message '[INFO] Writing surface model output to:  ' in routine LIS_writeModelOutput in file lis/core/LIS_historyMod.F90 produces garbage for all non masterproc processes.

Only the master process sets lsmoutfile.  Either all calls to LIS_writeModelOutput should initial the output file name to some clean value or only the master process should issue this message.

jvgeiger:
lislog.0000:
```
 [INFO] Writing surface model output to:
 OUTPUT/SURFACEMODEL/200210/LIS_HIST_200210290300.d01.gs4r
```

lisle.0001:
```
[INFO] Writing surface model output to:
^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@
```

PR #632, opened 2020-11-12
Fix HYMAP2 2-way coupling in parallel mode

dpsarmie:
HYMAP2 2-way coupling was not working properly when run in
parallel. This fix adds dummy subroutines to the land-surface
models which allow the 2-way coupling options to work properly.
A more robust fix might be added in the future.

Resolves: #605

Issue #605, opened 2020-10-09
Noah36 and HYMAP2 run issues


sujayvkumar:
Noah36 and HYMAP2 crashes with the following error: 
****************Error****************************
set sws routine for LSM + routing instance Noah-MP.4.0.1+HYMAP2 router is not defined
program will seg fault.....
****************Error****************************

PR #656, opened 2020-12-10
Fix garbage lislog output when writing model output

dmocko:
This commit fixes garbage lislog output for non-masterproc
log files when writing a log message about writing model
output.  These garbage log messages occur for surface, RTMs,
irrigation, and routing models.  This fix ensures that the
filename is written correctly before the log messages are
written.

Resolves: #409

### DIFF
--- a/lis/core/LIS_RTMMod.F90
+++ b/lis/core/LIS_RTMMod.F90
@@ -357,16 +357,17 @@ contains
                 open_stats = .false.
                 if(LIS_masterproc) then 
                    call LIS_create_output_directory('RTM')
-                   call LIS_create_output_filename(n,outfile,&
-                        model_name = 'RTM',&
-                        writeint=LIS_rtm_struc(n)%rtmoutInterval)
-                   if(LIS_rtm_struc(n)%stats_file_open) then 
+                   if (LIS_rtm_struc(n)%stats_file_open) then
                       call LIS_create_stats_filename(n,statsfile,'RTM')
                       LIS_rtm_struc(n)%stats_file_open = .false.
-                      open_stats = .true. 
+                      open_stats = .true.
                    endif
                 endif
-                
+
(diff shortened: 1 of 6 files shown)


bmcandr:
Hi @dmocko,
The changes look good. With regard to testing, I'm thinking of using the following testcases to test:

| Log output tested           | Testcase |
|----------------------------|----------|
| surface model + irrigation | noahmp3.6_gw |
| RTM                        | noahmp.3.6_cmem |
| HYMAP + NLDAS routers      | Public NLDAS2-e |
| HYMAP2                     | Testcase from open HYMAP2 PR |

Does that sound good?

dmocko:
Hi @bmcanrd - yes, that should cover all of the changes.  Thanks!

I don't expect any changes to the output files (just to the log files), but just to be safe, it might be worth comparing the output files for at least one of these tests.

David

bmcandr:
Hi @dmocko,

I successfully tested surface model, irrigation, RTM, and HYMAP2 logging, but I can't seem to get the HYMAP and NLDAS routing tests to run.

The HYMAP run fails with `set sws routine for LSM + routing instance Noah.3.6+HYMAP router is not defined` which makes sense as I don't see an equivalent of _noahmp36_getsws_hymap2.F90_ in the Noah.3.6 routing directory. Do you know how I might resolve this issue? (or is it worth the trouble given how simple this fix is?)

dmocko:
Thanks for testing @bmcandr .

I don't think it's worth the trouble.  We do need to decide as a team if we will keep supporting the old HyMAP router as well as the NLDAS router for all LSMs.

Looks like this was fixed for Noah-3.6 and HyMAP2 in issue #605 and fixed in PR #632.

David

bmcandr:
Thanks @dmocko , I'll go ahead and merge.

PR #662, opened 2020-12-11
Update devonly guards in Users' Guides for LISF 7.3 release

jvgeiger:
Update components of LISF that should be hidden within a development-only guard

This aspect of the Users' Guides should be removed, but a thorough code
review is required first.

### DIFF
--- a/docs/LIS_users_guide/LIS_usersguide.adoc
+++ b/docs/LIS_users_guide/LIS_usersguide.adoc
@@ -8,7 +8,7 @@
 :mathematical-format: svg
 :imagesoutdir: images
 :stem: latexmath
-:devonly!:
+:devonly:
 :lisfrevision: 7.3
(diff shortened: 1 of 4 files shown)


bmcandr:
Should the `devonly` attribute in _LIS_usersguide.adoc_ be set to false? Based on your comment it sounds like the devonly text should still be hidden so I just want to double-check before merging.

jvgeiger:
I am trying to set all the `devonly` attributes to true: 1)  to be consistent, 2) because that is how I have always done it.  I manually set that to false when I process the final documentation for the public release.

This was useful when we were using Subversion and had both a public branch and a development branch.  Now that we have just a master branch on git, this is no longer needed.  But before I remove it, I must determine whether the features hidden behind the `devonly` attribute made it into the master branch or they were dropped.

PR #670, opened 2020-12-16
Prevent divide-by-zero in Noah-MP with frozen soils

dmocko:
This pull request will prevent a divide-by-zero in the
Noah-MP-4.0.1 physics when removing sub-surface runoff.
This divide-by-zero is very rare, but can occur when
the soils are totally frozen.  The sub-surface runoff
will be zero as well as the hydraulic conductivity.
A simple check of the hydraulic conductivity was added
to prevent the divide-by-zero.

### DIFF
--- a/lis/surfacemodels/land/noahmp.4.0.1/phys/module_sf_noahmplsm_401.F90
+++ b/lis/surfacemodels/land/noahmp.4.0.1/phys/module_sf_noahmplsm_401.F90
@@ -7747,7 +7747,10 @@ ENDIF   ! CROPTYPE == 0
          END DO
 
          DO IZ = 1, NSOIL           ! Removing subsurface runoff
-         MLIQ(IZ) = MLIQ(IZ) - QDIS*DT*HK(IZ)*DZMM(IZ)/WTSUB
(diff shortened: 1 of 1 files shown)


bmcandr:
Hi @dmocko,

Is this related to the [recent ModelingGuru thread](https://modelingguru.nasa.gov/message/10725)?

I'm testing compilation as I'm guessing we don't have a testcase for this rare issue.

dmocko:
Hi @bmcandr - I think it might be related to that thread.  The user asked about a water imbalance issue, but I could see how a different set of compilation flags could end up with this issue triggering an imbalance - instead of triggering a floating invalid, which is what I was seeing.

This is specifically for the 557WW project, where a 12 member global ensemble simulation triggered this after 4+ years of simulation.  But I certainly knew where to look based on the user's comments on that thread.

I'm going to add to that MG thread and ask the user to just edit/add this fix in their branch, and see if the imbalance still occurs.

No, there's no testcase for this quick fix.

bmcandr:
Ah, I missed that this was for 557. I'll let @jvgeiger merge this to the support branch. Sorry for the confusion.

PR #682, opened 2021-01-04
Fix formatting of MODIS LAI options.

bmcandr:
Removed `devonly` block and fixed formatting.


dmocko:
Great - thanks for the quick fix.

PR #684, opened 2021-01-08
Create OBA output directories (for Air Force LISF 7.4)

emkemp:
This fixes a bug for writing Bratseth OBA files.  OBA files are written to different subdirectories, but in the prior code there was no way to *create* those subdirectories from LIS.  That bug has been fixed.


bmcandr (approved):

bmcandr:
Code looks good and produces no compiler errors. I will go ahead and merge.

PR #685, opened 2021-01-13
Change JULES-5.0 snow DA state variable quality control

dmocko:
This pull request includes changes to code that quality controls
JULES-5.0 state variables for snow DA for the SNODEP and USAFSI
products.  These changes include:

1) Turn off snow DA when snow depths are very high

2) Use snow density from the JULES-5.0 state variable

3) Check snow density against JULES-prescribed maximum/minimum

4) Only check SWE and snow depth maximum when doing an update

5) Increase the recommended SWE and snow depth maximum values
   in the attributes file


jvgeiger (commented):
This looks good to me.

dmocko:
Hi @yyoon4 and/or @yhkwon81 

Can you also please review these code changes?

I know we discussed them in our weekly tag-ups, but we'd like your concurrence before this is merged.

Thanks,

David

yhkwon81:
The changes look good to me, but I have a question. 

A more strict "deep snow" check was applied, as follows:

if (snod(t).gt.10.0) then
   update_flag(gid) = .false.
endif

However, the updated snow values are truncated, as follows:

if (swe(t).gt.swemax) then
   swe(t) = swemax
endif
if (snod(t).gt.snodmax) then
   snod(t) = snodmax
endif

Based on the code, deep snow (i.e., snow depth > 10 m) will not be updated. Then, the later code lines are just for the case when the abrupt update happens? Can I know the reason why snodmax was further increased (i.e., 100 m --> 1000 m), while much shallower snow depth criteria (i.e., 10 m) was used for turning off the update?

yyoon4:
@dmocko Thanks for your effort on this issues. Code changes look oK to me. Just minor comments.

--- comment 1
!A more strict "deep snow" check, independent of the ancillary data. - EMK
      if (snod(t).gt.10.0) then

Already @yhkwon81 mentioned that these hard codes restricted snow attribute table value (i.e., max for swe/snow depth)

To use attribute table, please consider these codes instead of hard code value. And if 10-m snow depth threshold is critical, please change the attribute value instead of hard coding. 

     if((snod(t).gt.snodmax) .or. swe(t).gt.swemax) then
        update_flag(gid) = .false.
     endif

--- comment 2
 #varmin  varmax	 		 #varmin  varmax	 
 SWE       		 SWE       
 0.0   10000.0		 0.0   100000.0
 Snowdepth 		 Snowdepth 
 0.0   100.0		 0.0   1000.0

-> Unit of SWE is mm and snow depth is m; so 10 m of snow depth is equivalent 10000 mm of SWE.

  --- comment 3
        if (swe(t).gt.swemax) then
            swe(t) = swemax
         endif
         if (snod(t).gt.snodmax) then
            snod(t) = snodmax
         endif

Not a big issue, but to keep consistency please consider below codes 

     if(swe(t).gt.swemax) then
        swe(t) = swemax
        snod(t) = swe(t)/sndens
     endif
     if(snod(t).gt.snodmax) then
        snod(t) = snodmax
        swe(t)  = snod(t)*sndens
     endif

     if(swe(t).lt.swemin) then
        swe(t) = swemin
        snod(t) = swe(t)/sndens
     endif
     if(snod(t).lt.snodmin) then
        snod(t) = snodmin
        swe(t)  = snod(t)*sndens
     endif

emkemp:
Hi @yhkwon81 

These changes were put in when running "warm-started" LIS-JULES for initializing GALWEM.  The "warm start" indicates initial conditions were provided by the UM operational model. We noticed that in glacier areas (Antarctica and Greenland), the existing swemax and snodmax attrib values would result in instantaneous removal of a large amount of snow mass provided from UM when model perturbations were turned on, even if DA was turned off.  Since that looked wrong, and since GALWEM would crash, I boosted the max values in the attrib files.

But LIS-JULES would, over a month of simulation, grow the snow depth and SWE to the new max values -- this was due to multiplicative model perturbations in the EnKF.  And GALWEM would still crash.  So, I finally added the 10-m strict check to disable updates for thick snow pack.  This stabilized the LIS-JULES snowpack growth over glaciers.

We could, as an experiment, rerun LIS-JULES with the original attrib max values and see what happens.

emkemp:
I submitted a test run to the debug QoS to see if the inflated max attribs are still needed.  Job is currently in the queue.

dmocko:
Thanks, @yyoon4, @yhkwon81, and @emkemp for your comments.

About Comment 1 - agreed.  I will use values from the attribute table instead of hard-coded, following the example.  I will run with the modified code first to see if a crash occurs.  If not, I will commit updated code.

About Comment 2:  I'm not sure why there are now four columns in the jules_snow_attribs.txt table.  Shouldn't there just be two?  Also, I understand that SWE is in mm and SnowDepth is in m.  However, due to the density of the snowpack, 10 m of snow depth is not equivalent to 10000 mm of SWE.  Eric is testing with the original attrib max values of 10000 mm for SWE and 100 m for snow depth - which are the same as used in Noah and Noah-MP.

About Comment 3 - agreed.  I will also change this code as part of my test for Comment 1.  If no crashes, I will commit an update.

David

yyoon4:
Hi @dmocko 

I am sorry for confusing you regarding comment 2.

I just copy and past your before and after codes, which makes confusing you. Just two attributes are correct!

And original attrib max values of 10000 mm for SWE and 100 m for snow depth will be correct. I am just confused about the density value.

10m (sd) = 10000mm (swe) / 1000 (max density);
100m = 10000mm / 100 (low density)

Thanks,
Yeosang

emkemp:
I finished a couple of 1-day runs, and I don't see any difference in the max SnowDepth or SWE between the original and new jules_snow_attribs.txt.  So, I am okay with keeping the original maximum values (same as Noah and NoahMP).

dmocko:
Hi all,

JULES kept running without any crashes over the weekend using the code that I just committed based on everyone's inputs.

Please review these latest codes and give approval or suggest changes.

Thanks,

David

yyoon4:
@dmocko Thanks for the update. The codes look great!

emkemp:
Code looks fine to me.

emkemp:
@yhkwon81  can you please review David's latest changes?  I want everyone to weigh in before we merge it in.

yhkwon81:
Hi,

David's latest changes also look good to me. Thanks a lot.

emkemp:
Excellent, thanks!

I will merge the code in.

PR #688, opened 2021-01-22
Fix garbage lislog output when writing model output

jvgeiger:
This commit fixes garbage lislog output for non-masterproc
log files when writing a log message about writing model
output.  These garbage log messages occur for surface, RTMs,
irrigation, and routing models.  This fix ensures that the
filename is written correctly before the log messages are
written.

Resolves: #409
(cherry picked from commit c785057346e8cfa4b43ecd1909a58e1cd95f96d9)

### DIFF
--- a/lis/core/LIS_RTMMod.F90
+++ b/lis/core/LIS_RTMMod.F90
@@ -357,16 +357,17 @@ contains
                 open_stats = .false.
                 if(LIS_masterproc) then 
                    call LIS_create_output_directory('RTM')
-                   call LIS_create_output_filename(n,outfile,&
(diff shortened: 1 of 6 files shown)


jvgeiger:
This is to facilitate debugging the support/lisf-557ww-7.3 branch.

dmocko:
Thanks @jvgeiger !

The code is identical to what I've manually edited and been running with for the spin-ups, so I am fine with merging this now.

David

emkemp:
On my end, I confirm the code compiles with Intel 19 w/o issue.

Issue #664, opened 2020-12-13
SWE Bug in LVT NLDAS-2 Reader (SAC)


jerlingi:
The SWE reader in LVT has a units mismatch when reading SAC data (missing a factor of 1000). No other NLDAS-2 models are affected. 

Files to be changed: LISF/lvt/datastreams/NLDAS2/readNLDAS2data.F90

@dmocko and I have tested a fix and will be committing it soon.

PR #158, opened 2019-06-20
Feature/GLASSalbedo_reader

abheeralisf:
The test-case computing the RMSE and BIAS between the GLASS-albedo and JULES-Albedo can be found in the following directory "/discover/nobackup/ahazra/UK_meto/JULES_LVT/example_08_sol/".

PR #689, opened 2021-01-25
Correct SAC SWE units

jerlingi:
This pull request addresses a units mismatch in the NLDAS-2 reader for Snow Water Equivalent (SWE) for the SAC land surface model. 

A test case can be found in /discover/nobackup/projects/wldas/share/sac_swe_testcase .

This fixes Issue #664 .

### DIFF
--- a/lvt/datastreams/NLDAS2/readNLDAS2data.F90
+++ b/lvt/datastreams/NLDAS2/readNLDAS2data.F90
@@ -575,6 +575,15 @@
               ((LVT_MOC_SWE(source).ge.1).or.                          &
                (LVT_MOC_TWS(source).ge.1))) then
              call retrieve_nldas2data(igrib,nc,nr,nvars,index,swe)
+             if (nldas2data(source)%lsm.eq."SAC") then !SAC has units of m 
+                do r = 1,nr
+                   do c = 1,nc
+                      if (swe(c+(r-1)*nc).ne.9999.0) then
+                         swe(c+(r-1)*nc) = swe(c+(r-1)*nc) * 1000.0
+                      endif
+                   enddo
+                enddo
+             endif
           endif
 
           if ((pid(index).eq.snowdepth_index).and.                     &


bmcandr:
Hi @jerlingi, can you open permissions on the testcase files? Thanks!

jerlingi:
Sorry! Should be fixed now!

dmocko:
Thanks @jerlingi for the bug fix!

For the benefit of others, we had previously discussed this bug fix, so I approve of the change (pending the successful testcase)

bmcandr:
@dmocko this PR introduces a fix for code in the 7.3 release so it should target the support/lisf-public-7.3 branch, correct?

dmocko:
@bmcandr Yes - it should go into the support/lisf-public-7.3 branch.  And then later merged into master.  Not significant enough of a bug to justify a new support/lisf-public-7.3 release, and can wait to be included whenever the next release is made.

bmcandr:
Hi @jerlingi, since this PR fixes a bug that exists in code that is already in the 7.3 release we're going to want this PR to target the support/lisf-public-7.3 branch rather than master. Unfortunately, simply editing the PR to target the support branch won't work because the other (~100) commits to master since the release will be swept up in the PR.

Follow these steps in your repo to move the fix to a new branch based off `support/lisf-public-7.3`:

1. `git switch -c bugfix/sac_swe_dup` -- create a duplicate of your current working branch
2. `git fetch upstream support/lisf-public-7.3` -- downloads the support branch, but doesn't merge it into your current branch
    * you may need to add NASA-LIS as an upstream first w/ `git remote add upstream https://github.com/NASA-LIS/LISF`
3. `git log --oneline -n 1` -- make note of the commit hash of the changes to relocate (*should* still be 1d2272c)
4. Confirm you're on the duplicate branch by checking `git status` (it should say `On branch bugfix/sac_swe_dup`)
5. `git branch -D bugfix/sac_swe` -- delete the original working branch; we're about to recreate it based off support
6. `git switch support/lisf-public-7.3` -- switch to the support branch
7. `git switch -c bugfix/sac_swe` -- create a branch with the original name, but based off of the support branch
8. `git cherry-pick bugfix/sac_swe_dup 1d2272c` -- cherry pick the changes in this commit to the fresh branch
9. Verify the change was committed to this branch (e.g., check `git log -n 1` and/or the contents of `readNLDAS2data.F90`)
10. `git push -f origin bugfix/sac_swe` -- push the branch up (we need to force push (-f) since we altered the history)
11. Click the *Edit* button at the top of the PR and change the target branch from  to `support/lisf-public-7.3`

Let me know if you have any questions or run into trouble!

jerlingi:
@bmcandr Thanks for the detailed instructions. I followed them, and it looks like everything worked.

bmcandr:
Hi @jerlingi, I was able to reproduce your output with the Intel19 compiler. With the GNU4 compiler on SLES11 I'm seeing differences in the _LVT_RMSE_FINAL.201710010000.d01.nc_ output in the _COUNT_SD_SWE_v_SWE_ and _COUNT_SD_SnowDepth_v_SnowDepth_ fields. For example, comparing the target to the output with `nccmp` yields:

```sh
DIFFER : VARIABLE : COUNT_SD_SnowDepth_v_SnowDepth : POSITION : [222,269] : VALUES : -9.71228e+08 <> 0
DIFFER : VARIABLE : COUNT_SD_SnowDepth_v_SnowDepth : POSITION : [222,270] : VALUES : -9.71228e+08 <> 0
DIFFER : VARIABLE : COUNT_SD_SnowDepth_v_SnowDepth : POSITION : [222,271] : VALUES : -9.71228e+08 <> 0
DIFFER : VARIABLE : COUNT_SD_SnowDepth_v_SnowDepth : POSITION : [222,272] : VALUES : -9.71228e+08 <> 0
DIFFER : VARIABLE : COUNT_SD_SnowDepth_v_SnowDepth : POSITION : [222,275] : VALUES : -9.71228e+08 <> 0
DIFFER : VARIABLE : COUNT_SD_SnowDepth_v_SnowDepth : POSITION : [222,276] : VALUES : -9.71228e+08 <> 0
DIFFER : VARIABLE : COUNT_SD_SWE_v_SWE : POSITION : [0,103] : VALUES : -9.71228e+08 <> 0
DIFFER : VARIABLE : COUNT_SD_SWE_v_SWE : POSITION : [0,104] : VALUES : -9.71228e+08 <> 0
DIFFER : VARIABLE : COUNT_SD_SWE_v_SWE : POSITION : [0,105] : VALUES : -9.71228e+08 <> 0
DIFFER : VARIABLE : COUNT_SD_SWE_v_SWE : POSITION : [0,106] : VALUES : -9.71228e+08 <> 0
DIFFER : VARIABLE : COUNT_SD_SWE_v_SWE : POSITION : [0,107] : VALUES : -9.71228e+08 <> 0
DIFFER : VARIABLE : COUNT_SD_SWE_v_SWE : POSITION : [0,108] : VALUES : -9.71228e+08 <> 0
```

The other files are identical, however. I wonder if this is similar to the issue I saw when reviewing #158 recently. Does the output for these fields in the target output look OK to you? If not, that will confirm there's a bug in the calculation of the COUNT_SD fields.

----

I also encountered the following warning message after the LVT run concluded with GNU4. @jvgeiger @dmocko have you seen this before? The lvtlog shows that the run finished.

```sh
--------------------------------------------------------------------------
mpirun has exited due to process rank 0 with PID 23503 on
node discover06 exiting improperly. There are three reasons this could occur:

1. this process did not call "init" before exiting, but others in
the job did. This can cause a job to hang indefinitely while it waits
for all processes to call "init". By rule, if one process calls "init",
then ALL processes must call "init" prior to termination.

2. this process called "init", but exited without calling "finalize".
By rule, all processes that call "init" MUST call "finalize" prior to
exiting or it will be considered an "abnormal termination"

3. this process called "MPI_Abort" or "orte_abort" and the mca parameter
orte_create_session_dirs is set to false. In this case, the run-time cannot
detect that the abort call was an abnormal termination. Hence, the only
error message you will receive is this one.

This may have caused other processes in the application to be
terminated by signals sent by mpirun (as reported here).

You can avoid this message by specifying -quiet on the mpirun command line.

--------------------------------------------------------------------------
```

bmcandr:
PS, you can take a look at my GNU4 output here: /discover/nobackup/projects/lis/PR_TESTCASES/689/testing_gnu4

dmocko:
Hi @bmcandr - Yes, it does seem similar to that previous PR.  I'm unclear what the "COUNT_SD" field represents, and why it's only in the RMSE output, and not in the other metrics.  There's nothing special in the METRICS.TBL file that would indicate that this field should be output for only the RMSE.

Additionally, I believe the correct output should be the fields with "0" from the GNU4 compiler, and not the garbage negative output from the Intel compiler.  This points to an initialization problems perhaps.  I bet that if this testcase is run in the Intel environment, but using the latest public support release (aka, not this PR), then we would also see this garbage output.

Finally - no, I haven't seen that warning message before using GNU4.  My experience with GNU4 is very limited.

jerlingi:
Hi @bmcandr, 

I checked the output in the GNU4 testcase, and if you view the LVT_MEAN_FINAL.201710010000.d01.nc, the variable SWE_from_SWE_v_SWE_ds1 (the SAC SWE) is still incorrect by a factor of 1000 (the original bug). I checked the /discover/nobackup/projects/lis/PR_TESTCASES/689/testing_i19 directory, and that looks reasonable in terms of the magnitude of the SWE variable in the LVT_MEAN_FINAL netCDF file.

bmcandr:
Hi @jerlingi, sorry for the confusion. After David commented I re-ran the testcase with an executable compiled from the support branch to verify that the issue with the COUNT_SD fields is unrelated to these changes. I will run the testcase with your fix again.

bmcandr:
I have re-run the testcase with GNU4. The fix output is in the _output/_ directory and the output of the base support branch is in _output_support/_. The COUNT_SD field in the _LVT_RMSE*_ from both runs is identical (zeros across the board) which implies the bunk data is due to a bug unrelated to these changes.

@jerlingi, I think we're OK to merge if the output in the other fields looks ok to you.

jerlingi:
Thanks for the clarification. Looks good to me now!

PR #697, opened 2021-01-29
Feature/crocus as sub model

LIS-navari:
This is a new development in the LIS framework. First, we implemented the Crocus Snow model (From SURFEX: https://www.umr-cnrm.fr/surfex/) as a stand-alone model into the LIS framework. Then we developed an infrastructure for coupling the Crocus with different LSM. As a pilot implementation, we coupled the Crocus with Noah-MP4.0.1. Crocus can be used both as a stand-alone model and a sub-model. 

Note that the model is under development.
 
Test case over the Colorado domain is available at: 
/discover/nobackup/mnavari/my_projects/thp/submodel/ssa_eval/test_for_GitHub

1- Generate the model parameter file using ldt.config.
2- Run the model using lis.config file


bmcandr:
Hi @LIS-navari,

Just wanted to give an update on this PR. I was initially seeing differences between our output. After trying a variety of different settings without success I realized that your executable may have been compiled without the `-fp-model precise` compiler flag that was merged last week. Sure enough, I was able to produce identical output on Intel 19 by compiling without that flag. I am currently testing on Intel 18 and GNU4 and will update you with the results tonight/tomorrow morning.

Meanwhile, please take a look at the results of my code review below. Thanks!

## Code review

@jvgeiger can you take a look at the following?:
* Changes to `lis/arch/Config.pl`
* `lis/make/Crocus.make`: is this the right place for this file? (see also GNU compile error caused by this file described below)

@LIS-navari:
* Document w/note about feature being under development:
  * Model attribute outputs in User Guide (in `lis/core/LIS_histDataMod.F90`)
  * New CROCUS config settings (`lis/surfacemodels/land/subLSM/crocus.8.1/Crocus81_readcrd.F90`)
* In `lis/core/LIS_soilsMod.F90`:
    * Re: the `if` block at lines 128-136 that is commented out. Do these changes allow the use of both soil texture and fraction? Just want to check that this error message can be safely removed.
* `lis/make/Crocus.make`: the compiler flags in this file are causing a compile failure with the GNU4 compiler:

    ```sh
    mpif90 -c -ffree-line-length-0 -O2 -fbacktrace -DSPMD -fconvert=big-endian -DHIDE_SHR_MSG -DNO_SHR_VMATH -DGFORTRAN -DLINUX  -I/discover/nobackup/projects/lis/libs/esmf/7.1.0r_gcc-4.9.2_openmpi-1.8.4/mod/modO/Linux.gfortran.64.openmpi.default -I/discover/nobackup/projects/lis/libs/ecCodes/2.7.0_gcc-4.9.2/include/ -I/discover/nobackup/projects/lis/libs/netcdf/4.5.0_gcc-4.9.2/include/ -I/discover/nobackup/projects/lis/libs/hdf5/1.10.1_gcc-4.9.2/include/ -DLIS_JULES -I./ -I../core/ -I../params/gfrac/SPORTDaily/ -I../params/gfrac/VIIRSDaily/ -I../surfacemodels/land/vic.4.1.1/ -I../surfacemodels/land/vic.4.1.1/physics/ -I../surfacemodels/land/vic.4.1.2.l/ -I../surfacemodels/land/vic.4.1.2.l/physics/ -I../surfacemodels/land/awral.6.0.0/physics/ -I../surfacemodels/land/rdhm.3.5.6/ -I../surfacemodels/land/rdhm.3.5.6/sac/ -I../surfacemodels/land/rdhm.3.5.6/frz/ -I../surfacemodels/land/rdhm.3.5.6/snow17/ -I../surfacemodels/land/rdhm.3.5.6/sachtet/ -I../surfacemodels/land/subLSM/crocus.8.1/phys/ ../surfacemodels/land/subLSM/crocus.8.1/phys/modd_isba_par.F90
    gfortran: error: unrecognized command line option ‘-autodouble’
    gfortran: error: unrecognized command line option ‘-noerror_limit’
    gfortran: error: unrecognized command line option ‘-auto’
    gfortran: error: unrecognized command line option ‘-WB’
    gfortran: error: unrecognized command line option ‘-traceback’
    make: *** [modd_const_atm.o] Error 1
    make: *** Waiting for unfinished jobs....
    gfortran: error: unrecognized command line option ‘-autodouble’
    gfortran: error: unrecognized command line option ‘-noerror_limit’
    gfortran: error: unrecognized command line option ‘-auto’
    gfortran: error: unrecognized command line option ‘-WB’
    gfortran: error: unrecognized command line option ‘-traceback’
    make: *** [modd_reprod_oper.o] Error 1
    gfortran: error: unrecognized command line option ‘-autodouble’
    gfortran: error: unrecognized command line option ‘-noerror_limit’
    gfortran: error: unrecognized command line option ‘-auto’
    gfortran: error: unrecognized command line option ‘-WB’
    gfortran: error: unrecognized command line option ‘-traceback’
    make: *** [modd_snow_par.o] Error 1
    ../dataassim/obs/SMAP_NRTsm/create_filelist.c: In function ‘create_filelist_’:
    ../dataassim/obs/SMAP_NRTsm/create_filelist.c:60:25: warning: passing argument 2 of ‘glob’ makes integer from pointer without a cast
        ierr = glob(regexp, NULL, NULL, &globbuf);
                            ^
    In file included from ../dataassim/obs/SMAP_NRTsm/create_filelist.c:15:0:
    /usr/include/glob.h:159:12: note: expected ‘int’ but argument is of type ‘void *’
    extern int glob (__const char *__restrict __pattern, int __flags,
                ^

    [ERR] Compile failed
    ```

* `lis/make/LIS_misc.h`: this file is compiled by the configure script and doesn't need to be committed. Probably was commited by accident!
* In `lis/surfacemodels/land/subLSM/crocus.8.1/Crocus81_coldstart.F90` @ line 90: change `MSG:` in log output to `[INFO]`?

bmcandr:
Hi @LIS-navari, I was able to produce identical output with the Intel 18 compiler as well so things look good.

WRT the GNU4 compile issue, Eric suggested during last week's tag up that we could also disable plugins that are still under development in _default.cfg_ until they are more stable. I tested this by disabling Crocus in the _user.cfg_ file and I did not hit the GNU4 compile error I mentioned above.

LIS-navari:
Hi @bmcanrd and @dmocko

In your first comment, you mentioned 

"In lis/core/LIS_soilsMod.F90:
Re: the if block at lines 128-136 that is commented out. Do these changes allow the use of both soil texture and soil fraction? Just want to check that this error message can be safely removed."

In NoahMP we need soil texture, and in Crocus, we need soil fraction information. 
To be able to use both datasets, I have commented out that section of lis/core/LIS_soilsMod.F90 subroutine. 

 @dmocko: I would be thankful if you could advise. 

Thanks

dmocko:
Hmmmm - from what I recall, this error message in LIS_soilMod.F90 (lines 128-136) was placed there to prevent confusion from users who might put both soil texture and the soil sand/clay/silt fractions into the LDT file, but then not be aware of their LSM would use the soil texture field or the soil fractions fields.

So, I don't think there is likely to be any harm in disabling this error message.  @LIS-navari is able to run with these lines commented out with Noah-MP.  But I can't be 100% sure right now, as I'd have to check every LSM.

Since this Crocus integration is still underdevelopment, can we put in an if block that skips these lines if one of the models is Crocus, but still does the check otherwise?

LIS-navari:
Thanks @dmocko 
I will add an if block to skip the soil texture and soil fraction check for Crocus.

LIS-navari:
Hi @dmocko @bmcandr 

I have added the following if block

+    logical :: SubModelIsCrocus

+       do i=1,LIS_rc%nSubLSMs
+          if (LIS_rc%subLSM(i).eq."Crocus8.1") then
+              SubModelIsCrocus = .true.
+              exit
+          endif
+       enddo 
+
+       if (.not. SubModelIsCrocus) then  
+         if(LIS_rc%usetexturemap(n).ne."none".and.&
+              LIS_rc%usesoilfractionmap(n).ne."none") then 
+
+            write(LIS_logunit,*) '[ERR] Please select either the soil texture or the soil '
+            write(LIS_logunit,*) '[ERR] fraction dataset. Both should not be enabled '
+            write(LIS_logunit,*) '[ERR] simultaneously ...'
+            call LIS_endrun()
+
+         endif
+       endif

Do you think this change would address your comment?
Thanks

LIS-navari:
Hi @dmocko @bmcandr 
@dmocko 
I have tried your suggestion; "can we put in an if block that skips these lines if one of the models is Crocus, but still does the check otherwise".
If I understand the process correctly, It does not work because we first initialize the domain (LIS_soils_init) and then initialize LSM (LIS_lsm_init()). 
The names of the model and sub-model are read during the initialization of LSM (LIS_lsm_init()). 
Do you have any thoughts on this?
Thanks

bmcandr:
Hi @LIS-navari, I stepped through the code in TotalView and you are correct that it does not seem possible to check the LSM/sub-LSM name during _LIS_soils_init_ execution.

@dmocko, what needs to be checked in the other LSMs? If we can confirm this change won't break anything, perhaps printing a [WARN] message when using both soil texture and soil fraction could replace this ERR message?

dmocko:
Sorry for not responding earlier.  Looking through the code, it doesn't appear that any checks about using both soil texture and soil fractions are done in the individual LSMs.  In fact, I think only Mosaic checks for the soil fractions, and I'm not sure about how well that model runs in the current code.

It looks like all of the checks for "usetexturemap" and "usesoilfractionmap" are in these three routines in LISF/lis/core:
- LIS_domainMod.F90
- LIS_readConfig.F90
- LIS_soils.F90

There are several if blocks that first check to see if the texturemap is used.  If it is, it never checks for the soil fraction maps:

       if(LIS_rc%usetexturemap(n).ne."none") then           
          call read_soiltexture(n)
       else
          if(LIS_rc%usesoilfractionmap(n).ne."none") then 
             call read_soilfraction(n)
          endif
       endif

That's why Mahdi is able to run with the ERR message commented out - it just does the first option, which is the texture map, if both are turned on.

The one exception to this may be in LIS_domainMod.F90, and the setting of flags "soilt_selected" and "soilf_selected".

I think changing to a warning message that emphatically states something similar to:  ...both texture and fraction selected.. both can't be used, so using soil texture and soil fractions will be ignored.

We should check that the internal LSM testcases will run as expected before this pull request is merged.

dmocko:
Reading my comment again, and my suggested warning message is confusing.  Here's another try:

[WARN] Both soil texture and soil fraction dataset are selected.
[WARN] Both should generally not be enabled simultaneously.
[WARN] For now, the soil texture will be used; the soil fraction
[WARN] will be ignored.  However, users should double-check
[WARN] the output of "Soiltype:" via the MODEL OUTPUT TBL.

David

sujayvkumar:
@dmocko @LIS-navari : I like this solution. Remember that we don't want to hard code model specific checks in the LIS core routines. As much as possible, avoid checks like 'if (LIS_rc%subLSM(i).eq."Crocus8.1") then' ..

jvgeiger:
I have pushed a few commits onto this pull request.

1. I removed the two additions to lis/arch/Config.pl.  These are hard-coded and specific to Crocus.  I moved them into Crocus.make.

Also, I believe that this was not working as intended.  The four files, ini_csts.F90, modd_surf_par.F90, modi_close_file.F90, and modi_get_luout.F90, that used either SFX_MNH or MNH_MPI_DOUBLE_PRECISION did not '#include "LIS_misc.h"', so both of these macros would then be treated as undefined.

2. Brendan, I worked with Mahdi on the lis/make/Crocus.make file.  The Crocus source files require specific compiler flags, so these flags were isolated into a Crocus-specific Makefile.

This will require further review to handle GNU compilers.

3. I disabled Crocus by default in the lis/make/default.cfg file.  Mahdi, you must create a lis/make/user.cfg file containing:

Crocus.8.1: On

The lis/make/user.cfg file should not be committed into the repository as it contains user-specified overrides.

4. I removed lis/make/LIS_misc.h.  As Brendan pointed out, it is temporary file created by the build scripts and was accidentally committed into the repository.

jvgeiger:
@LIS-navari, please review these files.  I believe that they do not need the '#include "LIS_misc.h"' statement:

Crocus81_coldstart.F90
Crocus81_main.F90
Crocus81_setup.F90
crocus_driver.F90

jvgeiger:
@LIS-navari, please update the copyright notice on our files.

LIS-navari:
Hi @jvgeiger @bmcandr 
I have applied the comments and updated the copyright notices. 
Please find a test case here:
/discover/nobackup/mnavari/my_projects/thp/submodel/ssa_eval/test_for_GitHub

bmcandr:
Thanks @lis-navari, the changes look good. I also successfully reproduced the testcase output with the Intel 19 compiler. Since this is feature is still in progress, we'll worry about creating an "official" testcase once it's polished up a bit and we're able to enable Crocus by default.

Per [David's comment above](https://github.com/NASA-LIS/LISF/pull/697#issuecomment-776313515), I'm going to run our internal LSM testcases to verify that the removal of the soil texture/fraction check doesn't break anything.

bmcandr (approved):

bmcandr:
The code looks good, the Crocus testcase produces identical results, and the LSM testcases all pass. I approve and will merge this to master.

Issue #705, opened 2021-02-05
LVT Max and Min metric time series output incorrect


dmocko:
The LVT "Max" and "Min" (maximum and minimum) metric time series output is incorrect.

The total analysis output of these metrics is correct (e.g., LVT_MAX_FINAL...  and LVT_MIN_FINAL...).

However, the time series output is incorrect (e.g., MAX_TS... and MIN_TS...).

It appears that the mean values are instead calculated and written to the max and min time series output.

This bug is in the support/lisf-public-7.3 branch, so will be including the fix into this branch.

I have a fix, and will be making a pull request with a testcase soon.

PR #706, opened 2021-02-05
Fix LVT Max and Min metric time series output

dmocko:
This pull request fixes time series output for the
Max (maximum) and Min (minimum) metrics from LVT.

The MEAN metrics code was cleaned-up as well, but
the output for this metric will not change at all.

A testcase is coming soon.

Resolves: #705

### DIFF
--- a/lvt/core/LVT_statsDataMod.F90
+++ b/lvt/core/LVT_statsDataMod.F90
@@ -40,6 +40,7 @@ module LVT_statsDataMod
 !
 ! !REVISION HISTORY:
 !  02 Oct 2008: Sujay Kumar; Initial version
+!  05 Feb 2021: David Mocko; Fixed Max and Min metric time series output
 ! 
 !EOP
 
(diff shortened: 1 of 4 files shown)


dmocko:
Testcase (with GrADS .xdf files and scripts) available here:
    /discover/nobackup/projects/lis/tools/pull_requests/706

bmcandr:
The changes look good and testing produced identical output to the target with Intel 19 and GNU4 compilers. I will go ahead and merge. Thanks @dmocko!

emkemp:
I was testing the same pull request, and noticed some differences in the MAX_SUMMARY_STATS.dat file.  I'll double check to see if this was just user error.

bmcandr:
Thanks for the heads up. Diff says the *_SUMMARY_STATS.dat are identical on my end.

emkemp:
I was getting zeros for the different regions (NWFRC, MBFRC, etc).  These are all defined by a Grads binary file, with the path given in TS_LOCATIONS_daily.TBL.

I compiled my LVT test with little endian.  I'm recompiling now with big endian to see if that was the issue.

emkemp:
It was the endianness flag.  When I switch to big endian, the results are identical.

No (retroactive) objections to this pull request.  :-)

dmocko:
Sorry for the trouble, @emkemp 

I shouldn't have included a path to a big_endian file in the TS_LOCATIONS file, especially: 1) without bundling up this file in the testcase tarball, and 2) not specifying to try this testcase with a particular endian-ness.

@bmcandr - If you will be adding this testcase to the suite, I would be glad to change my TS_LOCATIONS file so it doesn't use a binary file, and make new input/output testcase tarballs.  Just let me know.

PR #707, opened 2021-02-05
Support compiling LISF without HDF-EOS2

jvgeiger:
Several data assimilation obs readers require HDF-EOS2 support, but they
were checking for HDF4.  Thus if you disabled HDF-EOS2 at configure-time with

"Use HDFEOS? (1-yes, 0-no, default=1): 0"

then these routines would fail to compile.


emkemp:
Code looks fine, and compiles w/o issue.  I approve.

PR #725, opened 2021-02-25
Feature/smos nrt nn da

LIS-navari:
### Description
This is a new feature in the LDT and LIS to read  SMOS Level 2 Soil Moisture Near Real Time
Neural Network (L2-SM-NRT-NN) data product. 
LDT:
The new feature allows uses to read the SMOS data generate CDF, domain file, and DGG lookup table. 
LIS:
The new feature allows users to assimilate the SMOS data to different LSM. 

### Testcase
LDT: 
/discover/nobackup/mnavari/my_projects/smos/testcases/noah39_conus_10km/DA_proc_SMOS

LIS:
/discover/nobackup/mnavari/my_projects/smos/testcases/noah39_conus_10km/da


bmcandr:
Hi @LIS-navari, is this new feature meant for the 557WW LISF 7.4 release? If so, the pull request should target the *master* branch rather than *support/lisf-557ww-7.3* which is intended for fixes. You can edit the base branch at the top of the pull request to select *master*.

LIS-navari:
Hi @bmcandr   I have changed to master.

bmcandr:
Hi @LIS-navari, it's still showing support/lisf-557ww-7.3 for me. You might need to hit save after making the change.

jvgeiger (changes requested):
Hello @LIS_navari,


- ldt/DAobs/SMOS_NRTNN_L2sm/SMOSNRTNNL2sm_obsMod.F90
  - Document run-time options in ldt/configs/ldt.config.adoc.

- ldt/DAobs/SMOS_NRTNN_L2sm/write_lookup_table.F90
  - Why does LDT_SMOS_DGG_lookup need to `mkdir`?

- lis/dataassim/obs/SMOS_NRTNN_L2sm/SMOSNRTNNL2sm_Mod.F90
  - Document run-time options in lis/configs/lis.config.adoc.
  - Clean up SMOSNRTNNL2sm_setup by deleting "#if 0" and other commented out code.

- lis/dataassim/obs/SMOS_NRTNN_L2sm/read_SMOSNRTNNL2sm.F90
  - Use create_filelist function instead of system call to `ls`.
  - Why does this copy SMOS_filelist_sm.dat to SMOS_filelist_sm.????.dat?
  - See lis/dataassim/obs/NASA_SMAPsm/read_NASASMAPsm.F90
  - It seems that this routine creates only file_01.txt and file_02.txt.  The glob "file*.txt" may match too much.  How about:
    `call system (cat file_01.txt file_02.txt > ./SMOS_filelist_sm.dat')`
   `call system ('rm file_01.txt file_02.txt')`
  - Future enhancement (not for this PR): We should give the create_filelist function support to overwrite or append.

- lis/plugins/LIS_lsmda_pluginMod.F90
  - Plugins were registered for JULES 5.0, 5.2, and 5.3, but not for 5.x (5.5)
    Note that there is no support for JULES 5.1 nor JULES 5.4 in this plugin file.  I am not concerned about those two versions.

emkemp:
Thanks for looking at the code @jvgeiger.

On my end, I've compiled the existing pull request and found no problems with -3.  LDT also runs the test case w/o issue.  I'm having trouble getting LIS to run the test case, but this may be user error on my part.  I'll keep digging.

emkemp (approved):
I completed running the testcases using the original code in the pull request.  I successfully reproduced Mahdi's results.  Since then Mahdi has uploaded some code changes to address  review by @jvgeiger.  I see no code changes that would change results (removed code was either commented out or already disabled by #0), so the existing test results should be sufficient.

I approve the pull request.  An additional review by @jvgeiger is needed before merging.

LIS-navari:
Hi @jvgeiger @emkemp 

I have edited the code based on @jvgeiger comments except for this comment :
lis/dataassim/obs/SMOS_NRTNN_L2sm/read_SMOSNRTNNL2sm.F90

Use create_filelist function instead of system call to ls.
Why does this copy SMOS_filelist_sm.dat to SMOS_filelist_sm.????.dat?
See lis/dataassim/obs/NASA_SMAPsm/read_NASASMAPsm.F90
It seems that this routine creates only file_01.txt and file_02.txt. The glob "file*.txt" may match too much. How about:
call system ('cat file_01.txt file_02.txt > ./SMOS_filelist_sm.dat')
call system ('rm file_01.txt file_02.txt')
Future enhancement (not for this PR): We should give the create_filelist function support to overwrite or append.

I left this for Yonghwan because it is more complicated than I thought. We need to deal with files that the processing time and recording time are on two different days.

jvgeiger:
I am testing.

jvgeiger (approved):

jvgeiger:
By the way, I did test the update to create_filelist by running our NASA SMAP reader test, and I tested compiling with Jules 5.5.

jvgeiger:
@bmcandr, I am packaging this testcase.

PR #734, opened 2021-03-05
Reorganized read_CDFdata_month to reduce memory usage.

emkemp:
### Description

Running LIS-JULES in full Air Force configuration (10-km global domain,
12 ensembles, AGRMET Ops, land DA) leads to memory exhaustion on Discover
with 560 processors, when reading in the ASCAT CDF file.  These code changes
reduces the memory usage in read_CDFdata_month by about 50%, and LIS is able
to move past this stage.


jvgeiger:
Hello @emkemp, I ran a test, and I see differences that I did not expect.

Hopefully it is user error :-)

jvgeiger (approved):

PR #736, opened 2021-03-08
Fixed bug in writing RMSE STDEV counts to output.

### LINKED ISSUE #735: Fix output of RMSE sample size counts in LVT
### Bug Description

Before writing the sample sizes for the RMSE calculation, LVT will copy the data to a temporary array.  See lvt/metrics/LVT_RMSEMod.F90, subroutine LVT_writeMetric_RMSE.  Due to a typo, the temporary array is overwritten with data for a different metric:

                         value_total(:,m) = &
                              stats%rmse(m)%value_total(:,k,l)
                         count_value_total(:,m) = & 
                              stats%rmse(m)%count_value_total(:,k,l)
                         Value_ci(m) = stats%rmse(m)%value_ci(k,l)

                         value_stdev_total(:,m) = &
                              stats%rmse(m)%value_stdev_total(:,k,l)
                         count_value_total(:,m) = &   ! <-- EMK This is wrong
                              stats%rmse(m)%count_value_stdev_total(:,k,l)

Worse, the STDEV calculation is not supported in LVT restart files.  Together, if LVT is restarted, the sample size counts for RMSE will be smaller than they should be.

emkemp:
### Description

This fixes a bug in lvt/metrics/LVT_RMSEMod.F90, subroutine LVT_writeMetric_RMSE.  A temporary array containing sample sizes for the RMSE calculation is accidentally overwritten with data from another calculation (STDEV) before writing to file.  This latter calculation is not supported in the LVT restart file. So, if LVT is restarted, the sample size counts for RMSE will be lower than they should be.

The solution is to correct the name of the temporary array for STDEV sample sizes.

Resolves #735

### DIFF
--- a/lvt/metrics/LVT_RMSEMod.F90
+++ b/lvt/metrics/LVT_RMSEMod.F90
@@ -847,31 +847,33 @@ contains
   end subroutine computeSingleRMSE
 
 !BOP
-! 
+!
 ! !ROUTINE: LVT_writeMetric_RMSE
 ! \label{LVT_writeMetric_RMSE}
 !
-! !INTERFACE: 
+! !INTERFACE:
   subroutine LVT_writeMetric_RMSE(pass,final,vlevels,stats,obs)
-! 
+!
 ! !USES:
     use LVT_statsMod, only : LVT_writeSummaryStats
     use LVT_pluginIndices
 !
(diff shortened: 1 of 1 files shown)


bmcandr:
Tested with a testcase from an old PR and the output looks good. I'll go ahead and merge.

PR #742, opened 2021-03-09
Add modulefile for GNU 9.3.0/Open MPI 4.0.4 for discover

### LINKED ISSUE #741: Add GNU 9.3/Open MPI 4.0.4 development/running environment for discover
Add GNU 9.3/Open MPI 4.0.4 development/running environment for discover

jvgeiger:
### Description

Add GNU 9.3/Open MPI 4.0.4 development/running environment for discover

Important: I was unable to successfully install the HDF-EOS2 library.
You must must disable HDF-EOS2 support during configuration
(running ./configure):

```
Use HDFEOS? (1-yes, 0-no, default=1): 0
```

Resolves #741

### DIFF
--- a/env/discover/lisf_7_gnu_9.3.0_gmao_openmpi_4.0.4
+++ b/env/discover/lisf_7_gnu_9.3.0_gmao_openmpi_4.0.4
@@ -1,0 +1,166 @@
+#%Module1.0###################################################################
+
+proc ModulesHelp { } {
+    puts stderr "\t[module-info name] - loads the LISF_7_GNU_9_3_0_GMAO_OPENMPI_4_0_4 env"
+    puts stderr ""
+    puts stderr "This is for use on NCCS' discover system running SLES 12.3."
+    puts stderr ""
+    puts stderr "\tThe following env variables are set:"
+    puts stderr "\t\tDEV_ENV"
+    puts stderr "\t\tLIS_ARCH"
+    puts stderr "\t\tLIS_SPMD"
+    puts stderr "\t\tLIS_FC"
+    puts stderr "\t\tLIS_CC"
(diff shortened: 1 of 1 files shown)


bmcandr:
LIS also compiles slowly, but otherwise I am able to successfully compile each LISF component.

emkemp:
I'm of the opinion we should rewrite LVT_statsMod.F90 to avoid use of the giant Type of Types that causes gfortran to choke.  I acknowledge this is a gfortran bug rather than illegal Fortran; but if we're actually going to support gfortran, we should support compiling with optimization.

I'll take a quick stab at this, at least to start discussion.

emkemp:
This may be surprising easy to fix. 

I started moving LVT_statsEntry fields from the stats_struc data type, and turning individual entries into new variables. But, after looking in the LVT source code for specific invocations of swnet, lwnet, qle, etc, I don't see any.  What I do see is that the LVT_statsEntry type is designed to be an element in a linked list (the last field of LVT_statsEntry is a pointer to another LVT_statsEntry variable).  So, I don't think the LVT_statsEntry fields in stats_struc are actually used.

If I simply comment all of these fields out, LVT will compile.  No need for renaming.  I need to test run LVT to see what happens, but if I'm right, it will just work.

jvgeiger:
Yes, LVT still requires compiling with `-O0`.  I filled a bug report GCC Bugzilla, but it is still open.  I recall trying to refactor that module, but it either did not help or it was too much work.  So I decided to wait.  It sounds like you see something that I did not.  Good luck!

bmcandr:
Gotcha. Since the issue isn't with the modulefile, I'll go ahead and merge this PR. Thanks for looking into a fix for LVT, Eric!

PR #744, opened 2021-03-10
Fix slow/failed compilation of LVT by GFORTRAN

### LINKED ISSUE #132: LVT slow compiling with gfortran 7.3
The gfortran 7.3 compiler (used with the lis_7_gcc_7_3 private module on Discover) hangs in lvt/core/LVT_statsDataMod.F90, even with optimization turned off.

emkemp:
### Description

Newer versions of GFORTRAN will hang and often fail in lvt/core/LVT_statsDataMod.F90, due to an acknowledged compiler bug.  This pull request avoids the problem by commenting out all LVT_statsEntry fields in the LVT_stats object.

It turns out that these fields are not actually used by LVT -- the actual LVT_statsEntry objects are separately created and organized as a linked list.  Commenting out these fields does not change the runtime behavior of LVT, and allows compilation
with GFORTRAN.

Closes #132 

### Testcase

See /discover/nobackup/emkemp/bugfix/fix_lvt_gnu_compile/work, which runs LVT over CONUS with masking (using a little-endian binary file in common/).  The data compared are GDAS and NLDAS2. The common/ subdirectory includes a LDT parameter file for the domain, while data/ has symbolic links to GDAS and NLDAS2.  The test case runs for one month, and outputs some gridded metrics as well as summary (domain-wide) statistics.

before_code_change.METRICS_CONUS/ has output from LVT before changing the code.
after_code_change.METRICS_CONUS/ has output from LVT after changing the code.

The above runs were performed with module lisf_7_intel_19_1_3_304_traceback-work-around.  And I confirm LVT compiles with lisf_7_gnu_9.3.0_gmao_openmpi_4.0.4 (with optimization) after making the code changes.

(I don't know if this should go in support/lisf-public-7.3 or in master.  I chose the former.)


emkemp:
I just realized I forked from master.  So this can't go into support/lisf-public-7.3 -- it drags too many unrelated changes.

Let me know whether the actual fix should target support/lisf-public-7.3 or master, and I'll issue a new pull request.

bmcandr:
I vote for support/lisf-public-7.3 branch as well because it's a fix for released code.

PR #747, opened 2021-03-11
Commented out LVT_statsEntry fields in LVT_stats object.

### LINKED ISSUE #132: LVT slow compiling with gfortran 7.3

emkemp:
### Description

Newer versions of GFORTRAN will hang and often fail in
lvt/core/LVT_statsDataMod.F90, due to an acknowledged compiler bug.
This avoids the compiler bug by commenting out all LVT_statsEntry fields
in the LVT_stats object.

It turns out that these fields are not actually used by LVT -- the
LVT_statsEntry objects used instead are separately created and organized
as a linked list.  Commenting out these fields does not change the runtime
behavior of LVT, and allows compilation with GFORTRAN.

Closes #132  

### Testcase

See /discover/nobackup/emkemp/bugfix/fix_lvt_gnu_compile/work, which runs LVT over CONUS with masking (using a little-endian binary file in common/). The data compared are GDAS and NLDAS2. The common/ subdirectory includes a LDT parameter file for the domain, while data/ has symbolic links to GDAS and NLDAS2. The test case runs for one month, and outputs some gridded metrics as well as summary (domain-wide) statistics.

before_code_change.METRICS_CONUS/ has output from LVT before changing the code.
after_code_change.METRICS_CONUS/ has output from LVT after changing the code.

The above runs were performed with module lisf_7_intel_19_1_3_304_traceback-work-around. And I confirm LVT compiles with lisf_7_gnu_9.3.0_gmao_openmpi_4.0.4 (with optimization) after making the code changes.

### DIFF
--- a/lvt/core/LVT_statsDataMod.F90
+++ b/lvt/core/LVT_statsDataMod.F90
@@ -1275,190 +1275,190 @@ module LVT_statsDataMod
      integer          :: prev_mo_tavg
      integer          :: prev_yr_tavg
 
-     type(LVT_statsEntry) :: swnet        ! Net shortwave radiation (surface) (W/m2)
-     type(LVT_statsEntry) :: lwnet        ! Net longwave radiation (surface) (W/m2)
-     type(LVT_statsEntry) :: rnet         ! Net absorbed radiation (surface) (W/m2)
-     type(LVT_statsEntry) :: qle          ! Latent Heat Flux (W/m2)
-     type(LVT_statsEntry) :: qh           ! Sensible Heat Flux (W/m2)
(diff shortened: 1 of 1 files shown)


bmcandr (approved):

bmcandr:
I successfully reproduced the target output with both Intel 19 and GNU9 modulefiles. The output of the GNU9 run was not bitwise identical to the Intel19 target output as expected, but _very_ close ("identical" within a tolerance of 0.001).

I also confirm that LVT now compiles much faster (~1m35s) with -O2 optimization. I will go ahead and merge.

PR #759, opened 2021-04-09
Add AMSR2 snow depth DA to NoahMP.4.0.1

### LINKED ISSUE #758: Add AMSR2 snow depth DA to LIS-NoahMP.4.0.1
Adding the ability to assimilate AMSR2 snow depth observations with NoahMP.4.0.1 in FW branch. AMSR2 snow depths constrained by MOD10C1 collection

mlwrzesien:
### Description

LIS DA plugin updated to support AMSR2 snow depth DA and NoahMP401. GCOMW_AMSR2L3SND reader also updated to fix a unit mismatch between NoahMP401 and AMRS2. Use of MOD10C1 collection 5 has been replaced with collection 6 to support DA through the present.

Resolves #758 

### Testcase
Testcase available within EIS Freshwater project space
/discover/nobackup/project/eis_freshwater/mwrzesie/DA_AMSR2_snowdepth

### DIFF
--- a/lis/dataassim/obs/GCOMW_AMSR2L3SND/GCOMW_AMSR2L3SND_Mod.F90
+++ b/lis/dataassim/obs/GCOMW_AMSR2L3SND/GCOMW_AMSR2L3SND_Mod.F90
@@ -200,15 +200,19 @@ contains
     call ESMF_ConfigFindLabel(LIS_config,&
          "AMSR2(GCOMW) snow depth use MODIS (MOD10C1) data for snow detection:",&
          rc=status)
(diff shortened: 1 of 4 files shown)


bmcandr:
Hi @mlwrzesien, the code looks good and I'm able to reproduce your output so I will go ahead and merge. I'm including your note about these changes for context when we review changes for merger into the _master_ branch after the EIS pilot concludes:
 
>One concern is that I replaced the MOD10C1 collection 5 reader with one for MOD10C1 collection 6 within the AMSR2 code since collection 5 support ended a few years ago. however, I wasn't sure the best way to keep both C5 and C6 since I think it would have required an update to the lis.config file for AMSR2 DA.

PR #788, opened 2021-05-05
Feature/relax wb

### LINKED ISSUE #787: Turn off the water balance check in NoahMP.4.0.1
The water balance check in NoahMP 4.0.1 is too strict and is known to cause model run termination in several instances. This needs to be turned off (for the freshwater EIS project) to allow some of the DA updates to take shape.

sujayvkumar:
The water balance check (and the model termination steps) has been commented out. 
Fixes #787


sujayvkumar:
There is no test case associated with this pull request.

bmcandr:
Hi @sujayvkumar, do we want this change in master? The related issue suggests WB should only be turned off for the freshwater branch.

karsenau:
Hi @bmcandr,

Sujay may have a different thought on this update from the EIS branch, but I'm not sure if we want to remove this check?  Also, @dmocko - thoughts?

PR #798, opened 2021-05-18
Add support for multi-layer snow from JULES 5.0 (PS41 configuration)

emkemp:
### Description
Updates LISF to support multi-layer snow from JULES 5.0 (PS41 configuration)

Also adds preliminary support for NAVGEM SST data at 0.25 deg resolution.  Further work on this is expected.

Also updates LVT postprocessing scripts to use new module on DIscover, and to allow use of Skylake nodes.

This code was originally forked from support/lisf-557ww-7.3, but development lagged behind the creation of the support/lisf-557ww-7.4 branch.  Jim Geiger was instrumental in pulling the code out of the branch and rebasing it off of LISF 7.4.  

Code was tested by Eric Kemp.

PR #799, opened 2021-05-19
Added multi-layer snow variables for JULES 5.0 PS41 configuration.

emkemp:
### Description

This updates MODEL_OUTPUT_LIST.TBL.jules50 to add new variables associated with the JULES 5.0 multi-layer snow physics, activated with the PS41 configuration.


dmocko:
Hi @emkemp - does "SnowSoot:" need to also be turned on the MODEL TBL?  I don't see this variable added in many of the templates and Python scripts in PR #798 .

emkemp:
Hi @dmocko:

SnowSoot: can be turned OFF for PS41, because the variable is always set to zero.  But it's possible a future physics configuration will generate useful information for that variable, which is why we added it as an output variable.

Issue #825, opened 2021-07-06
Add documentation on CONTRIBUTING a Discussion, Issue, or Pull Request


dmocko:
The LIS team is migrating the use of Modeling Guru for questions to a new Discussions tab on GitHub.

Documentation on how to best contribute, and to help users decide whether to open a Discussion, Issue, and/or Pull Request will be added to the repository.

Contributions to this documentation will be welcome.

dmocko:
Added a draft pull request.

I welcome any improvements/additions.

We should add a short section about a testcase being required for a new feature (but not a bug fix).

Any thoughts about adding a "Code of Conduct?".   This seems like a great resource:  https://www.contributor-covenant.org/

PR #826, opened 2021-07-06
Create CONTRIBUTING.md

dmocko:
Added a first draft of the CONTRIBUTING document to the LISF repository

### Description

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,49 @@
+# How to contribute
+We are glad you are reading this.  Thanks for taking the time to contribute!
+
+## Resources
+* LIS website: https://lis.gsfc.nasa.gov/
+* [How to read LISF documentation](https://github.com/NASA-LIS/LISF/tree/master/docs)
+* [LDT Users' Guide](https://github.com/NASA-LIS/LISF/blob/master/docs/LDT_users_guide/LDT_usersguide.adoc)
+* [LIS Users' Guide](https://github.com/NASA-LIS/LISF/blob/master/docs/LIS_users_guide/LIS_usersguide.adoc)
+* [LVT Users' Guide](https://github.com/NASA-LIS/LISF/blob/master/docs/LVT_users_guide/LVT_usersguide.adoc)
+* [Working with GitHub](https://github.com/NASA-LIS/LISF/blob/master/docs/working_with_github/working_with_github.adoc) for LISF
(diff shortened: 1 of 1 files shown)


dmocko:
This is still a draft.

See Issue #825

bmcandr:
This is looking great @dmocko! I just made a few minor edits described in my commit message.

I wonder if we should also ask contributors to update the docs if they're adding new config options?

dmocko:
Thanks @bmcandr for the improvements.

Yes, we should specify that users will need to update the appropriate UG if new config options are added.  Great reminder.

I also think that we should indicate that users agree that their contributions will be under the terms of the LIS license:
https://github.com/NASA-LIS/LISF/blob/master/LICENSE.txt

dmocko:
I added a reminder about the Users' Guide(s) and new/modified config entries - and added in a part about the license.

I also added the testcase requirement in a previous commit.

bmcandr:
Just took another pass over this to restructure and simplify it a bit while also adding more details here and there.

The use of questions for the section headers made it tricky to link to each section in the ToC so I simplified them to a single word (e.g., "I am having a problem running LDT/LIS/LVT and would like to ask a question" became "Support"). I'm not married to this change so if we feel the questions are better, we can revert the headers.

Open to edits and suggestions!

dmocko:
Hi @bmcandr 

Great - thanks so much for the changes.  I think it's clearer and the section headers will help repeat visitors to this document re-find whatever is needed.

I agree - I will remove the "draft" label, and am fine with this being reviewed one last time before merging in.

David

bmcandr:
I made changes to `docs/README.md` yesterday in #860 that update the instructions for viewing the Users' Guides and add links to each guide. Therefore, I decided to remove the links to the individual Users' Guides here as they would take users to the unrendered *.adoc file. Instead, this doc will take users to the README which will then point them to the Users' Guides in a readable format (if they follow the instructions 😄).

That said, I'm thinking the documentation section here (or in docs/README?) could be beefed up:
* provide an example of the the config calls that need to be documented if added/modified (e.g., `call ESMF_ConfigFindLabel`)
* describe where else new LSMs and metforcing readers should be documented (e.g., `doc s/LIS_users_guide/user_cfg_table.adoc` and `obtain_source.adoc`?)

These could come in a subsequent PR.

bmcandr:
Merging this for now.

Issue #843, opened 2021-07-19
GCC 10 incompatability


pratiman-91:
### Bug Description
Unable to compile lis and ldt (did not try with lvt but possibly affected) with GCC 10.
```bash
gfortran -c -pass-exit-codes -ffree-line-length-0 -O2   -fconvert=big-endian -DGFORTRAN -I/opt/lisf_libs/mod/modg/Linux.gfortran.64.mpiuni.default/  -I/opt/lisf_libs//include/ -I/opt/lisf_libs//include/  -I/opt/lisf_libs//include/  -I./ -I../lib/xmrg/ -I../params/VIC/ -I../USAFSI/ ../core/LDT_gfracMod.F90
../core/LDT_gfracMod.F90:478:65:

  478 |                        n, LDT_gfrac_struc(n)%gfrac%value(:,:,k) )
      |                                                                 1
Error: Missing actual argument for argument ‘_formal_8’ at (1)
make: *** [Makefile:108: LDT_gfracMod.o] Error 1

[ERR] Compile failed
```

### Steps to Replicate

Compile LIS and LDT using gcc (Ubuntu 10.3.0-1ubuntu1) 10.3.0 and GNU Fortran (Ubuntu 10.3.0-1ubuntu1) 10.3.0

### Solution

Add flag ```-fallow-argument-mismatch``` to lis and ldt compile scripts.
Additionaly add flag ```-fno-allow-invalid-boz``` for ldt. Due to errors in ```readSMOPSsmObs.F90```
```bash
gfortran -c -pass-exit-codes -ffree-line-length-0 -O2   -fconvert=big-endian -DGFORTRAN -I/opt/lisf_libs/mod/modg/Linux.gfortran.64.mpiuni.default/  -I/opt/lisf_libs//include/ -I/opt/lisf_libs//include/  -I/opt/lisf_libs//include/  -I./ -I../lib/xmrg/ -I../params/VIC/ -I../USAFSI/ ../DAobs/SMOPS/readSMOPSsmObs.F90
```

emkemp:
Thank you for this bug report.  I successfully replicated the LDT compilation error with gfortran 10.1.0 on the NASA DIscover supercomputer running SLES 12.3.  I will investigate further.

emkemp:
Update:  The LDT compilation error also occurs with gfortran 11.1.0.

pratiman-91:
Thanks @emkemp .

[...]

PR #850, opened 2021-07-21
LDT updates to allow compilation with gfortran 10+

emkemp:
### Description

Issue #843 discusses a number of syntax errors found by gfortran 10 and 11, which were reproduced on Discover.  Generally the bugs fall into three categories:
* At least one instance where an optional argument is used by a Fortran subroutine that does not have an interface (explicit, or created via a module); this is illegal in ISO Fortran.
* The readgfrac and readlaisai function calls are made with inconsistent argument lists, which causes newer gfortran to give a cryptic syntax error.
* The use of binary integer constants with parameters in readSMOPSsmObs is illegal under ISO Fortran; instead these constants must be used in DATA statements.

The fixes fall into five categories:

* Update gfrac and lai/sai Fortran read codes to use the same number of arguments (2).
* Update the gfrac and LAI/SAI C codes to remove unnecessary arguments.
* Update FLAKE code to move read_FLake_lakedepth into FLAKE_paramsMod, allowing use of optional argument.
* Modified readSMOPSsmObs to move usage of binary integer constants to DATA statements.
* Minor editorial changes along the way.

Test compilations were made with GFORTRAN 10 and 11, plus existing official modules lisf_7_intel_19_1_3_304 and lisf_7_gnu_9.3.0_gmao_openmpi_4.0.4.  No runtime tests were performed.

Note:  Issue #843 also implicates LIS (and possibly LVT), but this pull request is restricted to LDT.

### Testcase
Recommend testing this code with existing unit test suite.


emkemp:
I just realized this fix is oriented toward master.  Should I redo this for support/lisf-public-7.3?

emkemp:
I'm replacing this pull request with PR #851, which targets support/lisf-public-7.3.

PR #870, opened 2021-08-02
Updated Working With GitHub Document to replace URL syntax

emkemp:
Most of the git instructions for interacting with GitHub still used the https
URL syntax, which will stop working in late Aug 2021.  The revised text
replaces the syntax with the new "github:username/repo" syntax to be used
with ssh keys.  Two important exceptions:

* URL syntax is preserved for instructions that require using a web page (e.g.,
  first cloning the NASA-LIS LISF.git repo to another GitHub acccount).
* URL syntax is preserved when the context refers to "old" syntax and settings
  that should be updated.

Updated .adoc file was tested with asciidoctor 2.0.15 (installed via miniconda3), 
and HTML output was visualized with Firefox installed on Discover.

### DIFF
--- a/docs/working_with_github/working_with_github.adoc
+++ b/docs/working_with_github/working_with_github.adoc
@@ -1,7 +1,7 @@
 = Working with GitHub
 :author: James Geiger
-:revnumber: 12
-:revdate: 20 May 2021
+:revnumber: 13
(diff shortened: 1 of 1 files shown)


bmcandr:
Thanks for making this update @emkemp! Looks good on my end.

PR #873, opened 2021-08-03
Updated LISF development workflow diagram

emkemp:
### Description

Replaced hand-drawn diagram of LISF development workflow in git.
New diagram was created on app.diagrams.net, saved as a .drawio file,
and converted to png.

Revised working_with_github.adoc file was processed with
asciidoctor, and the resulting html file was inspected with
Firefox on Discover.


bmcandr:
I like the minimal design @emkemp! I actually started working on something similar a while back, but it's much busier (and uglier) than yours...

LISF-git-workflow

Are we able to make edits to your version? I have a few suggestions:

1. Borrowing from my version, it might be helpful to indicate where pull requests are occurring.
2. Re-label "Release Branch" as "Support Branch" to align with our branch names.
3. Delete the arrow from fix branch to master and add an arrow from the last commit on the support branch to the last commit on master. This reflects that the fix is merged into support via PR then support is merged back into master.

Lastly, do we need the LISFLOW.io file in the PR since we have the PNG file?

emkemp:
Hi @bmcandr:

Yes, we can edit the diagram by loading the LISFLOW.drawio file back into app.diagrams.net (that's why I kept it). I agree with your edit suggestions and will incorporate.

emkemp:
Latest diagram version is ready for review.

Also, I vote to keep the .drawio file, so we can update the diagram later if we change our development workflow.

bmcandr:
Thanks @emkemp. Looks good to me.

>Also, I vote to keep the .drawio file, so we can update the diagram later if we change our development workflow.

Agreed.

PR #893, opened 2021-09-03
Support compiling with Cray compilers

jvgeiger:
### Description

These updates provide support for compiling LISF on Cray systems using Cray's compilers.

### DIFF
--- a/ldt/ANNdata/GCOMW_AMSR2_TB/GCOMWAMSR2TB_ANNdataMod.F90
+++ b/ldt/ANNdata/GCOMW_AMSR2_TB/GCOMWAMSR2TB_ANNdataMod.F90
@@ -61,7 +61,6 @@ contains
 ! !INTERFACE: 
   subroutine GCOMWAMSR2TB_ANNdatainit()
 ! !USES: 
-    use ESMF
     use LDT_coreMod
     use LDT_timeMgrMod
     use LDT_logMod
--- a/ldt/ANNdata/GHCNsnwd/GHCN_ANNdataMod.F90
+++ b/ldt/ANNdata/GHCNsnwd/GHCN_ANNdataMod.F90
@@ -63,7 +63,6 @@ contains
 ! !INTERFACE: 
   subroutine GHCN_ANNdatainit()
 ! !USES: 
-    use ESMF
     use LDT_coreMod
     use LDT_timeMgrMod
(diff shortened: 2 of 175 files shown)


emkemp:
I've started test compiles of this pull request on Discover to make sure nothing was broken.  I'll do Intel 19, then gfortran 11.

emkemp:
Intel 19 compiles fine on Discover, with both optimization level 2 and -3 (except for a known bug in JULES which is detected in the latter case).

I'm testing gfortran 11 now.

emkemp:
Update:  gfortran 11 won't compile this code, because my gfortran related pull requests have not been accepted yet (see #851, #854, and #855).  I will try using an older gfortran instead.

emkemp:
GFORTRAN 9 complained about the use of single quotes in an #include statement in the AGRMET code, so I replaced with double quotes.  This update has been pushed back to the pull request.  No other issues noted with LDT or LIS.  T

For LVT, GFORTRAN 9 hangs for a few minutes when compiling LVT_statsDataMod.F90.  The problem is the use of unnecessary entries in the stats_struc type.  This problem was fixed in master via PR #747, and when I manually repeat the fix in the pull request code, LVT compiles even with optimization.  @jvgeiger We should discuss how to migrate this change from master to the present PR, as this may also fix the problem you reported with the Cray compiler.

jvgeiger:
I successfully shuffled commit 0db7eef07d57f83e61cb5a09d72295eb3193ac45 (PR #747) around.

I am now testing on narwhal.

jvgeiger:
LVT must still be compiled with -O0 when using the Cray compilers.  The Cray compilers hang on lvt/core/LVT_LISoutputHandlerMod.F90.  The Cray compilers worked on this one file for 25 minutes, then I gave up.

I am able to compile the rest of LVT with -O2 using the Cray compilers.

jvgeiger:
lvt/core/LVT_LISoutputHandlerMod.F90 can be compiled with -O1 using the Cray compilers.

emkemp:
Latest code compiles on Discover w/o issue with Intel 19 and GCC 9.2.  LDT will not compile with gfortran 11.  In the interest of time I will accept this pull request, but more work is needed if we want to support newer GCC versions.  See pull requests  #851, #854, and #855.

PR #913, opened 2021-10-18
Feature/vis-portal

sahmad3:
### Description

Update EIS visualization portal directory with files from Freshwater, Sea Level Rise, and Fire pilots.


dmocko:
Hi @sahmad3 - thanks for all this work and the PR for the eis-freshwater branch.

I won't be qualified to do a full review of this PR, but I would recommend that the .DS_Store files are not committed/merged.

bmcandr (approved):
Approving this for now. We may need to make some changes to the content of the `eis_freshwater_visualization_portal/` directory as we review and polish up the eis_freshwater branch for merger into master

PR #926, opened 2021-11-09
Patches to S2S scripts for Nov 2021 Air Force Deliverable

emkemp:
### Description

Numerous changes to S2S scripts for November 2021 Air Force deliverable, including replacing most shell scripts with Python 3.


jvgeiger:
Thanks.

jvgeiger:
1) I would like to remove the `#SBATCH --mail-user`  lines from the scripts.  I realize that this is a placeholder for the user to fill in, but if the user does not then

* `#SBATCH --mail-user=` will end up not sending an email notification
* `#SBATCH --mail-user=user@email.com` will end up sending an email to email.com, which is a registered domain and which most likely will bound back to SLURM.

Setting `#SBATCH --mail-type=END` is sufficient to tell SLURM to send an email notification to the submitter upon completion of the job.


2) The lvt/utils/usaf/s2smetric/README files refers to "8u" scripts.   The document calls out 8 scripts, but there are 9 in that directory - make_s2s_metric_geotiff.py is not referred to.  Should I correct this to 8 or 9?

karsenau:
Thanks @jvgeiger!

@karsenau can take on #1, if @emkemp wants to do #2

emkemp:
I'll tackle the README file.  (Sorry about that!)

emkemp:
I moved make_s2s_metric_geotiff.py into a separate directory.  It is only in the repository in case we really need to create separate GeoTIFFs for each metric ensemble member (no sense throwing it away).  But is it not considered part of s2smetric at this time.

The README has been updated.

emkemp:
I removed the --mail-user= in the script for the Bratseth retuning software (separate from S2S, won't interfere with the updates from @karsenau)

karsenau:
Thanks @emkemp !

I'm just about to commit the other removed "mail" lines in the slurm scripts.

karsenau:
@jvgeiger and @emkemp:

I realized the following script:  
generate_s2safricom_da_config_scriptfiles_fcst.py

Has sort of a "misleading" naming convention, as this is the script that generates the forecast run, not the da run. If it's ok with you all, I would like to modify the script name and update some of the descriptions in the script.

The file currently exists in:
lis/configs/557WW-7.4-FOC/S2S_AFRICOM/generate_s2safricom_da_config_scriptfiles_fcst.py 

Thanks!

emkemp:
Discover is going down at 9 am for maintenance, so you may be out of time.  I would push your mail-user changes now.

karsenau:
@emkemp :  Commit pushed over

emkemp:
Got it!  @jvgeiger I think the fixes are all in.

PR #931, opened 2021-11-16
Add Discover Quick Start Guide

bmcandr:
### Description

Adds a Quick Start guide for new LISF users on Discover. This document is a compilation of guidance shared with me by Jim, Eric, David, and others when I joined the LIS team (as well as some things I picked up along the way.)

This document can be viewed on GitHub directly: https://github.com/bmcandr/LISF/blob/feature/discover-quick-start-guide/docs/discover_quick_start/discover_quick_start.adoc

Or as rendered AsciiDoc: https://raw.githubusercontent.com/bmcandr/LISF/feature/discover-quick-start-guide/docs/discover_quick_start/discover_quick_start.adoc

**I welcome any additions or suggested edits!**

### DIFF
--- a/docs/README.adoc
+++ b/docs/README.adoc
@@ -27,6 +27,7 @@ AsciiDoc easily converts to HTML; there are even Firefox and Chrome plugins that
 * https://raw.githubusercontent.com/NASA-LIS/LISF/master/docs/LVT_users_guide/LVT_usersguide.adoc[LVT Users' Guide]
 
 .Additional Guides:
+* https://raw.githubusercontent.com/NASA-LIS/LISF/master/docs/discover_quick_start/discover_quick_start.adoc[Quick Start Guide for NCCS Discover Users]
 * https://raw.githubusercontent.com/NASA-LIS/LISF/master/docs/working_with_github/working_with_github.adoc[Working with GitHub]
 * https://raw.githubusercontent.com/NASA-LIS/LISF/master/docs/public_testcase_walkthrough/public_testcase_walkthrough.adoc[Public Testcase Walkthrough]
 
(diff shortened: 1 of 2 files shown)


jvgeiger (commented):
I suggest that the clean _.profile_ file should be

```
# This file is read each time a login shell is started.
if [ -n "$PS1" ]; then
    echo "" ; echo "sourcing clean .profile" ; echo ""
fi

module use --append $HOME/privatemodules

ulimit -s unlimited

if [ -n "$PS1" ]; then
    echo ""
    echo "sourced clean .profile"
    echo "--------------------"
    echo ""
fi
```

It should not load any modules.

I recommend that, when you are compiling and/or running LISF, you load only the lisf_7_intel_19_1_3_304 modulefile - do not load any other modules.

When you load too many unrelated modules into your environment, you run the risk of creating undetected incompatibilites between tools and libraries. For example, you may end up loading three different NetCDF libraries. And one day LIS runs, but the next day LIS cannot read its own restart file.

You can point readers to our _Creating a custom modulefile_ document, in docs/creating_a_custom_modulefile/creating-a-custom-modulefile.adoc, for more information.

bmcandr:
Thanks for the feedback @jvgeiger. I'll make the changes tomorrow.

jvgeiger:
I still say that we should not encourage people to load modules in their _.profile_ files.  I spent too much time cleaning up people's environments.  Moving module commands out of the _.profile_ files was the primary reason for creating our custom modulefiles. :-)

We should remind everyone that the lisf_7_intel_19_1_3_304 module file is for compiling, running, and debugging LISF.

bmcandr:
D'oh. I thought I had removed that per your suggestion. I'll make the fix.

bmcandr:
Should be good now @jvgeiger!

bmcandr:
Beat me to that last commit :)

jvgeiger:
Does my commit look good to you? Do you have anything else to add?

bmcandr:
Just pushed another change to more explicitly discourage users from loading multiple modules alongside the LISF module. I think that should be it from me now. Thanks for the help!

jvgeiger:
This Quick Start guide looks good.  This will be very useful.  Thanks!

Issue #954, opened 2021-12-10
Noah-MP-4.0.1 LSM crashes when running multiple nests


dmocko:
When running Noah-MP-4.0.1 LSM, LIS will crash when running multiple nests.

LIS crashes because it tries to allocate variables on the 2nd nest that are already allocated on the 1st nest.

This crash does not occur when running a single nest.

I have a solution, and will soon be making a pull request.

This fix will go on the lisf-public-7.3 branch.

PR #955, opened 2021-12-10
Deallocate Noah-MP-4.0.1 variables in coldstart for multiple nests

dmocko:
This bug fix deallocates variables within the NoahMP401_coldstart
routine, to prevent a LIS crash when running with multiple nests.

Resolves: #954


emkemp:
I don't think this branch is supposed to be in the NASA-LIS repo.  It should be in the dmocko repo.

@jvgeiger , will there be a problem if I accept this pull request?

dmocko:
@emkemp - good catch!  I will close this PR and open a new one.

PR #958, opened 2021-12-15
Fix/retune bratseth

emkemp:
### Description

Bug fixes for LIS 7.4 Air Force branch, involving Bratseth error covariance retuning software.

* Changed lis/compile script to update path to Fortran code, so retuning executables are actually compiled.
* Updated batch script for running on Discover (new environment module)
* Updates to disable GEOPRECIP, SSMI, and CMORPH processing
* Logic change to first guess settings to prevent Scipy from issuing a fatal error for wind speed best-fit (not sure if this is due to Scipy version change on Discover)
* Tightened the screening for 10-m wind speed and 2-meter temperature to reduce scatter in recent 28-day empirical semivariograms (created for 29 Nov through 5 Dec).

### DIFF
--- a/lis/compile
+++ b/lis/compile
@@ -116,7 +116,7 @@ fi
 # Compile standalone Fortran utilities for tuning Air Force Bratseth scheme.
 if [ "$compile_autotune" = "YES" ]; then
     echo "Compiling Air Force Bratseth tuning utilities"
-    if cd utils/afwa/retune_bratseth/src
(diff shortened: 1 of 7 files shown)


jvgeiger (commented):
In blacklist_spd10m.cfg and blacklist_t2m.cfg, it looks like omb_thresh was changed from an integer to a float.  Is that OK?

jvgeiger:
Sorry.  I misread the diffs.  Looks good.

PR #967, opened 2022-01-12
Add workaround for Cray compilers

jvgeiger:
### Description

The Cray compilers have an issue compiling
jules.5.0/src/io/model_interface/model_interface_mod.F90
when using either the -O2 or -O1 optimization flags.  Force this file
to compile with -O0.


emkemp:
Code works on Discover, so nothing broke.  I approve.

PR #973, opened 2022-01-28
Inherit max path length from ESMF

bmcandr:
### Description

Replaces the arbitrary 500 char max path length added in #944 with ESMF's built-in `ESMF_MAXPATHLEN` which holds the value of 1024.

An alternative approach would involve abandoning `LIS_CONST_PATH_LEN` altogether and replacing it with `ESMF_MAXPATHLEN`.

### Testcase

Tested with the MERRA-2 testcase from our internal suite by modifying the `MERRA2 forcing directory` entry to artificially create a long path, for example:
`MERRA2 forcing directory: ././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././input/METFORCING/MERRA2`

### DIFF
--- a/lis/core/LIS_constantsMod.F90
+++ b/lis/core/LIS_constantsMod.F90
@@ -21,11 +21,12 @@ module LIS_constantsMod
 !
 !EOP
 !BOC
+   use ESMF, only : ESMF_MAXPATHLEN
    public
 !----------------------------------------------------------------------------
 ! software constants
 !----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


bmcandr:
I'm going to abandon `LIS_CONST_PATH_LEN` and use `ESMF_MAXPATHLEN` instead. I'm closing this and will open a new PR after making the change to the metforcing readers I've already updated.

bmcandr:
After talking with Jim, we're going to stick with our internal constant and set it with ESMF's value. Resurrecting this PR accordingly. 🧟

Issue #1001, opened 2022-03-08
Modify GEFS reader for 0.25 degree data


dpsarmie:
The current reader is only able to read in 0.50 degree GEFS data. Modifications should be made to the reader to allow the reader to ingest 0.25 degree data.

PR #1002, opened 2022-03-08
Modify GEFS reader to support 0.25 deg data

dpsarmie:
This commit will add support for the 0.25 deg GEFS data. The reader modifications
will keep the 0.50 deg functionality and add a config option to allow the user
to switch between 0.50 deg and 0.25 deg GEFS data. There is a bug with the
reforecast data when reading in the GEFS wind fields and this will be fixed in a
subsequent fix.

Resolves: #1001 

### Testcase
Testcase can be found here: /discover/nobackup/dpsarmie/work/gitLIS/GEFS_025Reader/testcase


bmcandr:
Hi @dpsarmie, overall this PR looks good and I am able to reproduce your target output. I have one recommendation: add a check for the valid resolutions (either 0.25 or 0.5) when the `"GEFS forecast grid resolution:"` entry is present in the lis.config file. In the current implementation the resolution defaults to 0.5 deg if this entry is not present, but accepts any value if the entry is present. See this excerpt from the NLDAS2 readcrd file as an example:

```
  call ESMF_ConfigFindLabel(LIS_config,"NLDAS2 data center source:",rc=rc)
  call LIS_verify(rc, 'NLDAS2 data center source: not defined')
  do n=1,LIS_rc%nnest
     call ESMF_ConfigGetAttribute(LIS_config,nldas2_struc(n)%nldas2_filesrc,&
          default="NCEP",rc=rc)
     if ((trim(nldas2_struc(n)%nldas2_filesrc).ne."GES-DISC").and.   &
         (trim(nldas2_struc(n)%nldas2_filesrc).ne."NCEP")) then
        write(LIS_logunit,*) '[ERR] NLDAS2 data center source:'
        write(LIS_logunit,*) '[ERR] must be equal to either "GES-DISC"'
        write(LIS_logunit,*) '[ERR] or to "NCEP".  Please check your'
        write(LIS_logunit,*) '[ERR] config file.  Stopping....'
        call LIS_endrun()
     endif
  enddo
```

dpsarmie:
Ok, I've added a check for the invalid values and kept the functionality of defaulting to 0.50 when the option is not in the config file.

bmcandr (approved):

Issue #1011, opened 2022-03-09
LAI from the lis_input file is not being used in Noah-MP-4.0.1


dmocko:
LAI from the lis_input file is not being used in Noah-MP-4.0.1 LSM for some of the dynamic vegetation physics options.

Options 7, 8, or 9 are listed as "input LAI"; however, the LAI from the lis_input file is currently not being sent to the Noah-MP-4.0.1 physics when these option numbers are provided.

Other options for "dynamic vegetation option" (options 1 through 6) are not affected by this bug.

This bug was noted and a solution provided in Discussion #992.

A pull request to the public-7.3 support branch is forthcoming.

PR #1013, opened 2022-03-10
Fix use of input LAI for Noah-MP-4.0.1 vegetation options

dmocko:
For Noah-MP-4.0.1 dynamic vegetation options 7, 8, or 9, the input
LAI was not being correctly used in the model physics.  This pull
request fixes this bug, and allows the use of input LAI or SAI in
the model physics.  The LIS Users' Guide was also updated to give
further details of the config entries that need to be provided for
all of the different options.

Options 1 to 6 for this physics option is not affected by this fix.

Resolves: #1011

### DIFF
--- a/lis/surfacemodels/land/noahmp.4.0.1/NoahMP401_main.F90
+++ b/lis/surfacemodels/land/noahmp.4.0.1/NoahMP401_main.F90
@@ -511,6 +511,8 @@ subroutine NoahMP401_main(n)
 ! fill in the LAI or SAI values from MPTABLE.
                if (tmp_lai.eq.LIS_rc%udef) then
                   tmp_lai          = NOAHMP401_struc(n)%noahmp401(t)%lai
+               endif
+               if (tmp_sai.eq.LIS_rc%udef) then
                   tmp_sai          = NOAHMP401_struc(n)%noahmp401(t)%sai
                endif
             else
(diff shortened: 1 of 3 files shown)


lis/surfacemodels/land/noahmp.4.0.1/NoahMP401_main.F90
```diff
@@ -483,8 +485,38 @@ subroutine NoahMP401_main(n)
...
+! input LAI vegetation options, fill in the SAI values from MPTABLE.
+               if (LIS_rc%useSAImap(n).ne."none") then
+                  tmp_sai          = LIS_sai(n)%tsai(tid)
+               endif
+! If any LAI or SAI values are undefined at a tile,
+! fill in the LAI or SAI values from MPTABLE.
+               if (tmp_lai.eq.LIS_rc%udef) then
+                  tmp_lai          = NOAHMP401_struc(n)%noahmp401(t)%lai
+                  tmp_sai          = NOAHMP401_struc(n)%noahmp401(t)%sai
+               endif
```
bmcandr, line 515: This only checks LAI for an undefined value. Should SAI also be checked? Or is it assumed that if LAI is undefined, so is SAI?

dmocko: Good catch.  I just pushed another commit to resolve this, after testing the change.

PR #1018, opened 2022-03-17
Updates the H2 offline capability

### LINKED ISSUE #1017: Enhance the offline running mode with HyMAP
When HyMAP runs by reading existing LIS outputs, allow the use of data that may not be co-located with the same domain.

sujayvkumar:
Allows H2 to run with LIS outputs that are not necessarily colocated with the LIS domain.
The outputs can be from a larger domain with different spatial resolution

Resolves #1017

### Description

### DIFF
--- a/lis/configs/lis.config.adoc
+++ b/lis/configs/lis.config.adoc
@@ -10140,10 +10140,14 @@ where LIS outputs are located.
 `LIS runoff output interval:` specifies 
 LIS output time step
 
+`LIS runoff output domain file:` specifies location of the LIS runoff output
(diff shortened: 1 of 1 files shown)


sujayvkumar:
A testcase is at : /discover/nobackup/projects/HMIDAS/svkumar/NOSL/Tests/H2offlinetest2

bmcandr:
@sujayvkumar, looks like I don't have access to the HMIDAS project directory so I can't access the testcase.

bmcandr (approved):

PR #1019, opened 2022-03-17
Include LIS endrun statement for missing USAF files

karsenau:
The USAF/AGRMET retrospective forcing reader does
 not stop running when an input forcing file is missing.
 In that case, same forcing values could be interpolated
 farther into time or 0s could be interpolated in, when
 no values present.

 Added a LIS_endrun call to the routine along with
 when it was called with an "[ERR]" message and applied
 "[WARN]" to other warning statments when other errors
 occur during a file read. (Note: Reader continues
 with no stopping when the latter occurs).

### Description

### DIFF
--- a/lis/metforcing/usaf/readagrmetforcinganalysis.F90
+++ b/lis/metforcing/usaf/readagrmetforcinganalysis.F90
@@ -26,7 +26,8 @@ subroutine readagrmetforcinganalysis(n,findex, order, agrfile, month)
 ! !USES:
   use LIS_coreMod, only         : LIS_rc, LIS_domain, LIS_masterproc
   use LIS_timeMgrMod,only       : LIS_get_julhr,LIS_tick,LIS_time2date
-  use LIS_logMod, only          : LIS_logunit, LIS_verify, LIS_warning
+  use LIS_logMod, only          : LIS_logunit, LIS_verify, LIS_warning,&
+                                  LIS_endrun
   use LIS_spatialDownscalingMod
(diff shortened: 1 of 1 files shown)


emkemp:
It looks like there are other scenarios in that subroutine where the program returns without reading data from the AGRMET file.  In effect, the same bad behavior still exists, but with different triggers (e.g., file permission issues).

I'm looking at lines 111-116, lines 123-135, 169-179, and 287-296.  Maybe more LIS_endrun calls should be inserted?

karsenau:
I had wondered that too and wasn't sure if those "warnings" were deliberately that way for a reason (unknown)?

I would agree that LIS_endruns could be inserted where a failed read occurs, and I can add if that seems like an agreeable way to go for the team.

karsenau:
@jvgeiger -- any thoughts on this?

jvgeiger:
Hello,

I agree that those warnings should become errors with a call to LIS_endrun.

Also, ferror is never used, so we should remove it.

karsenau:
Thanks @jvgeiger for your feedback on those conditional block updates in the AGRMET/USAF retrospective forcing reader.

I'll make the updates and submit by tomorrow morning.

karsenau:
@jvgeiger - Pushed over updated commit in response to your inputs.

https://github.com/NASA-LIS/LISF/pull/1019/commits/cd387201233a6f35617cc7667df2d564726ac54b

Please let me know if you need anything else.

Issue #1022, opened 2022-03-23
Grid not set correctly in GEFS Reader


dpsarmie:
The GEFS reader should set the grid spacing to the user defined variable instead of being hard-coded. 

The following variables: gefs_struc(n)%gridDesc(9) and gefs_struc(n)%gridDesc(10) should be changed in gefs_forcingMod.F90 to use the config defined values.

PR #1023, opened 2022-03-23
Fix GEFS grid input

dpsarmie:
The grid spacing variables gefs_struc(n)%gridDesc(9) and gefs_struc(n)%gridDesc(10) were not being set according to the config option. This has now been fixed for both the reforecast and operational dataset.

The old test case that was previously submitted should still work to test this commit if needed.

Resolves: #1022

### DIFF
--- a/lis/configs/lis.config.adoc
+++ b/lis/configs/lis.config.adoc
@@ -6055,8 +6055,15 @@ or mean sea level (msl) can be selected (which is not supported yet).
 of GEFS ensemble members being read in. Reforecast2 currently
 supports up to a 11 members.
 
-`GEFS forecast grid resolution:` specifies the grid resolution of the GEFS
-forecast data. The current resolutions supported are 0.50 and 0.25.
+`GEFS forecast grid resolution:` specifies the grid resolution in degrees
+of the GEFS forecast data. The current resolutions supported are:
+
+|====
+|GEFS forecast type | Resolutions
+
+|`Reforecast2`      | `0.25` and `1.0`
+|`Operational`      | `0.25` and `0.50`
+|====
 
 .Example _lis.config_ entry
 ....
--- a/lis/metforcing/gefs/gefs_forcingMod.F90
+++ b/lis/metforcing/gefs/gefs_forcingMod.F90
@@ -216,8 +216,8 @@ contains
               gefs_struc(n)%gridDesc(7) = -90.0   ! original GEFS
               gefs_struc(n)%gridDesc(8) = lonEnd
 !              gefs_struc(n)%gridDesc(8) = 359.00  ! original GEFS
-              gefs_struc(n)%gridDesc(9) =  1.0
-              gefs_struc(n)%gridDesc(10) = 1.0
+              gefs_struc(n)%gridDesc(9) =  gefs_struc(n)%gefs_res
+              gefs_struc(n)%gridDesc(10) = gefs_struc(n)%gefs_res
               gefs_struc(n)%gridDesc(20) = 64  ! -180 to 180?
 !              gefs_struc(n)%gridDesc(20) = 0   ! for 0 to 360?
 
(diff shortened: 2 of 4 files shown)


bmcandr:
Hi @dpsarmie, I'm seeing differences in the output when using the 0.25 deg testcase. That should be expected given the change made here, correct?

dpsarmie:
Hi @bmcandr , that's correct. There shouldn't be a difference in the 0.50 deg testcase (I'm assuming there is one?) but there should be a difference in the 0.25 deg case, which is now correct.

I can replicate the 0.25 deg case on my end if needed.

bmcandr:
Thanks. I have a 0.5 deg testcase I can use.

bmcandr:
@dpsarmie, do you mind reproducing outputs for the 0.25 deg testcase? I'm seeing differences with the 0.5 deg testcase that I will need to investigate (it is comparatively pretty old). If you want to take a look my working directory is `/discover/nobackup/projects/lis/PR_TESTCASES/1023/testing_05deg`

dpsarmie:
Haha, so I was actually just about to shoot @karsenau an email about an issue that is coincidentally affecting your test case replication.

The GEFS reader runs in either with either operational or reforecast data. The operational data on Discover is at 0.50 degrees but it looks like the reforecast was at 1.0 deg? (@karsenau can you confirm?)

The test case is for Reforecast data, which is at 1.0 degree. So gefs_struc(n)%gridDesc(9) should be 1.0 and can't be set to gefs_struc(n)%gefs_res because that's at 0.50 deg via the config file.

If that's the case, then I'll need to go back and add checks and include 1.0 as a valid option (in the readcrd file) and figure out the handling because we would now have GEFS Reforecast data at 1.0 degree and 0.25 degree and GEFS Operational data at 0.50deg and 0.25deg on Discover.

karsenau:
Hi @dpsarmie and @bmcandr:

Yes, the original GEFS reader was set up and tested 3-years ago with the original GEFS Reforecast (v2) "latlon" dataset at 1.0 deg resolution.  

I don't know if anyone will use the older GEFS "Reforecast v2" dataset, so if you want to not include it any more (or place it as a "if 0 / end" directive around that part of the file griddesc domain entries for now with a note for the older dataset), then that shouldn't impact the use of the lis config entry using resolution.

dpsarmie:
Ok, thanks @karsenau.

@bmcandr, your call on the way forward. I'm ok with phasing out the 1.0deg reader and leaving in some legacy code with "if 0 / end" in the reader. If not, then it should be viable to keep the functionality in the reader with some changes.

I also have the 0.25 deg Reforecast2 reader pretty much ready to go for a separate pull request (was trying to make different commits for the 0.25 operational and reforecast reader). I can include a separate test case for that reader if needed.

Let me know what you think makes sense.

bmcandr:
Thanks for all the info. I'll follow up tomorrow with a suggested course of action.

bmcandr:
Ok, we can proceed with the changes on the master branch as [the public support branch only supports the 1 deg reforecast dataset](https://github.com/NASA-LIS/LISF/blob/21d623e0bcb317034037a85b546e04b466014df8/lis/metforcing/gefs/gefs_forcingMod.F90#L193-L221).

I'm leaning towards keeping support for the 1deg because it may require fewer changes. Fixing support for the different resolutions looks like it should only require some additional checks in readcrd_gefs.F90 to make sure that a valid resolution is selected for the forecast type:

https://github.com/NASA-LIS/LISF/blob/e760240d71be25791949533b4a1d059b7d25e9e4/lis/metforcing/gefs/readcrd_gefs.F90#L70-L91

gefs_forcingMod.F90 already seems to handle the different resolutions.

What do you think?

dpsarmie:
Sounds good. I can add checks for the following cases:
Reforecast2 mode: 1.0deg or 0.25 deg
Operational mode: 0.5deg or 0.25 deg

Any other combinations will call the endrun function. Want me to place the modifications in this PR or should I open a separate pull request?

bmcandr:
This PR is fine since it rounds out the fix in the initial commit.

dpsarmie:
Ok, this should be good to go now.

bmcandr:
Thanks. Testing now.

bmcandr:
Testing looks good, but I think we need one more check to set the default resolution based on forecast type. Currently, if no resolution is specified it defaults to 0.5 degrees which is only supported by the Operational forecast type. I'm going to update the config doc so I can add this check as well.

bmcandr:
@dpsarmie, I reproduced the target outputs for the 0.25 and 1deg tests with changes I just pushed. Feel free to merge if you approve.

dpsarmie:
Hi @bmcandr, these changes look good. Your edits should already be reflected on my branch for this PR, correct? Or did you mean something else with 'merge if I approve'?

PR #1042, opened 2022-04-05
Feature/li siodist

sujayvkumar:
### Description

A new output format called 'distributed binary' was added. In this format, each processor writes outputs for the respective local domain to a binary file. At the output time steps, no synchronization is done to gather output variables across multiple processors. 

The quilting of the outputs across multiple processors is then done using LVT using a new run mode called "LIS post processing"

Resolves #1039 

### Testcase

A testcase can be found at : /discover/nobackup/projects/eis_freshwater/svkumar/Tests/LIS_test_distio

the lisrun directory contains the LIS example config. In this setup 16 processors should be used. 
the lvtrun directory then contains the configuration to quilt the files into a single file in netcdf format. 

The outputs will be generated in the same directory as that of the LIS output.

PR #1049, opened 2022-04-07
Write soil layer global attributes to quilted NetCDF file in LVT "LIS postprocessing" mode

### LINKED ISSUE #1047: LVT's LIS Postprocessing does not write soil layer attributes to quilted NetCDF
### Bug Description

The new LIS Postprocessing mode introduced to LVT does not write soil layer global attributes to the quilted NetCDF file. Related to #1042.

I have a fix ready and will open a PR soon.

bmcandr:
### Description

The "LIS postprocessing" mode introduced to LVT in #1042 does not write `NUM_SOIL_LAYERS` or `SOIL_LAYER_THICKNESSES` global attributes to the quilted NetCDF file because they are not contained in the LIS-generated binary output files. This fix adds optional LVT configuration entries for both attributes so users can write these attributes if they wish.

The `history` attribute is still not written. I'm torn on whether to add support for that too. It does [*not* appear to be required by the CF conventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#description-of-file-contents) and users would have to enter, and update, a properly formatted string in their config file (e.g., `"created on date: 2020-07-17T11:04:15.799"`). Seems prone to error and inaccuracy.

@sujayvkumar, let me know what you think.

Resolves #1047

### Testcase

`/discover/nobackup/projects/lis/PR_TESTCASES/1042/testing/`

### DIFF
--- a/lvt/configs/lvt.config.adoc
+++ b/lvt/configs/lvt.config.adoc
@@ -2632,3 +2632,23 @@ Number of fields in the LIS output: 11
 ....
 Number of vertical levels in the LIS output: 4
 ....
+
+`LIS output number of soil moisture layers:` specifies the number of soil moisture layers used in the LIS output. This is optional and defaults to 0.
+
+`LIS output soil moisture layer thickness:` specifies the number of soil moisture layers used in the LIS output. These values will be written to the NetCDF as global attributes and should match those used in the LIS configuration file to define the LSM soil moisture layer thicknesses. If the number of soil moisture layers is omitted or set to 0, this entry is not read and soil moisture layer thicknesses are not written to the NetCDF global attributes.
+
+.Example _lvt.config_ entry
+....
+LIS output number of soil moisture layers:  4
+LIS output soil moisture layer thickness: 0.1 0.3 0.6 1.0
+....
+
+`LIS output number of soil temperature layers:` specifies the thicknesses of soil moisture layers in the LIS output. This is optional and defaults to 0.
+
+`LIS output soil temperature layer thickness:` specifies the thicknesses of soil temperature layers in the LIS output. These values will be written to the NetCDF as global attributes and should match those used in the LIS configuration file to define the LSM soil temperature layer thicknesses. If the number of soil temperature layers is omitted or set to 0, this entry is not read and soil temperature layer thicknesses are not written to the NetCDF global attributes.
+
+.Example _lvt.config_ entry
+....
+LIS output number of soil temperature layers:  4
+LIS output soil temperature layer thickness:  0.1 0.3 0.6 1.0
+....
--- a/lvt/core/LVT_LISpostMod.F90
+++ b/lvt/core/LVT_LISpostMod.F90
@@ -62,6 +62,8 @@ module LVT_LISpostMod
      integer   :: npes
      integer   :: nfields
      integer   :: nlevels
+     integer   :: nsoilmoistlayers, nsoiltemplayers
+     real, pointer :: soilmoistlyrthk(:), soiltemplyrthk(:)
   end type lispost_type_dec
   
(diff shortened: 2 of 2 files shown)


sujayvkumar:
@bmcandr  - The soil layer options can be added as user specific config options in LVT. Once you obtain them, we can add to the attributes. 

I think we can let go of the history attribute.

sujayvkumar:
@bmcandr  : Looks good. Thanks Brendan!

dmocko:
I agree that the history attribute is not needed at this time, but if/when we ever get the point where LVT can generate GES DISC-ready LDAS netCDF files, this attribute will be needed.  We will need to add it back in, along with many other attributes.

bmcandr:
@dmocko, that link indicates that they are not actually required by the CF convention.

>For backwards compatibility with COARDS **none** of these global attributes is required.

dmocko:
@bmcandr,  Good catch - thanks for the clarification.

Yes, probably explains the differences, and I was more following requests from the GES DISC (who were telling me some attributes were for CF compliance) instead of a super-strict read of the text at that link.

So, it's required/requested from the GES DISC, but not for CF compliance.

dmocko:
One other thing about the LDAS datasets at the GES DISC:  Instead of variables with multiple layers (such as soil moisture), which requires a global attribute about listing the soil layer thicknesses - the GES DISC requested each soil layer as a separate variable.

```
        float SoilM_0_10cm(time, lat, lon) ;
        float SoilM_10_40cm(time, lat, lon) ;
        float SoilM_40_100cm(time, lat, lon) ;
        float SoilM_100_200cm(time, lat, lon) ;
```

The layer structure is in the variable name.  This also makes it a little easier to handle if the LSM has a different layer structure for soil moisture layers and soil temperature layers.

bmcandr:
I updated my fix to use the existing LVT config entries for soil moisture and also added support for soil temperature layer attributes. I modified the attribute names to indicate the difference so `NUM_SOIL_LAYERS` is now `NUM_SOIL_MOIST_LAYERS` and `NUM_SOIL_TEMP_LAYERS`, respectively. Same goes for the thickness attributes.

I've added a testcase as well at this location: `/discover/nobackup/projects/lis/PR_TESTCASES/1042/testing/lisrun_clean`

emkemp:
I dusted off this pull request.  The new soil layer thickness output from the test case is in meters, but other parts of LIS and LVT treat this as centimeters.  Do we want to add another option to specify the units in the LVT output file for clarification?

emkemp:
I'm rescinding my earlier comment.  It turns out the soil layer thicknesses are set in the lvt.config, so it is trivial to change the thicknesses from meters to centimeters.

The larger question of whether units should be added as an additional attribute can be handled by a separate pull request.

@dmocko Do you believe the GES DISC requirement of separate 2d arrays for soil fields for each soil layer is important enough to add to this pull request, or can it be addressed separately?

emkemp:
*PING*

Hi @dmocko:

Do you believe the GES DISC requirement of separate 2d arrays for soil fields for each layer is important enough to add to this pull request, or can it be addressed separately?

dmocko:
Hi @emkemp 

The GES DISC request for separate 2d arrays for soil fields for each layer can be addressed in a separate pull request.

I am in favor of having LIS directly output the soil fields as separate 2nd arrays (as a config option) instead of in LVT.

emkemp:
Thanks @dmocko.   I will coordinate with @jvgeiger on handling the test case before approving this pull request.

cmclaug2:
I confirm that the `NUM_SOIL_MOIST_LAYERS` and `NUM_SOIL_TEMP_LAYERS` attributes correctly define the number of soil layers.  I will approve this PR.

PR #1055, opened 2022-04-14
Added -fp0 flag for Cray compilers, for IEEE compliance.

emkemp:
### Description

Added fp0 to optimization flag list for Cray compiler for LDT.

This flag appears to yield the closest conformance to the IEEE floating point standard for the Cray compiler, and helps
improve agreement in USAFSI analyses produced on HPC11 versus analyses produced on Discover with the -fp-model consistent flag for the Intel compiler.

Similar changes may be needed for LIS and LVT, but have not been tested yet.  Adding -fp-model consistent for Intel compilers is also deferred at least until @jvgeiger returns from leave.

### DIFF
--- a/ldt/arch/Config.pl
+++ b/ldt/arch/Config.pl
@@ -175,8 +175,9 @@ elsif($opt_lev == 1) {
    $sys_c_opt = "";
 }
 elsif($opt_lev == 2) {
-   if($sys_arch eq "cray_cray") {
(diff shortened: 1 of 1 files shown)


emkemp:
Since this compiler flag change is needed immediately for the Air Force project, and since @jvgeiger is on leave and cannot approve, I will accept my own pull request and stoically accept the consequences.

bmcandr:
Accept your own pull request?

emkemp:
...

PR #1058, opened 2022-04-15
Fix: enable LDT to compile without HDF5

### LINKED ISSUE #1057: LDT fails to compile in USAFSI code without HDF5
### Bug Description

If LDT is configured without HDF5 support, compilation fails in two USAFSI modules: `USAFSI_amsr2Mod.F90` and `USAFSI_xcalgmiMod.F90`. HDF5 is an optional dependency in according to our user guides so LDT should be able to at least compile without it.

See #1056 for error messages and discussion.

The following fixes should be merged into the `support/lisf-557ww-7.4` branch:

* In `USAFSI/USAFSI_amsr2Mod.F90` add `#if (defined USE_HDF5)` to line 218 and `#endif` to line 631 (just before `end subroutine read_amsr2_attributes`)
* In `USAFSI_xcalgmiMod.F90` add `#if (defined USE_HDF5)` to line 236 and `#endif` to line 507 (just before `end subroutine read_xcalgmi_attributes`)

### Steps to Replicate

Configure LDT with HDF5 disabled, then compile.

bmcandr:
### Description

Adds checks for HDF5 usage to allow LDT to compile without HDF5.

Resolves #1057

### Testcase
Configure LDT with HDF5 disabled and then compile.

### DIFF
--- a/ldt/USAFSI/USAFSI_amsr2Mod.F90
+++ b/ldt/USAFSI/USAFSI_amsr2Mod.F90
@@ -238,6 +238,7 @@ contains
                                                   tb23h_raw(:,:), tb23v_raw(:,:), &
                                                   tb37h_raw(:,:), tb37v_raw(:,:), &
                                                   tb89h_raw(:,:), tb89v_raw(:,:)
+#if (defined USE_HDF5)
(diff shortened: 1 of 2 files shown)


dmocko:
Thanks @bmcandr for the quick fix.  I approve and will close/merge.

PR #1069, opened 2022-04-20
Updated Cray compiler flags for LIS with -O2.

emkemp:
### Description

Updated -O2 compiler flags for Cray compiler for LIS.


emkemp:
In the interest of time, I will merge my own pull request so I can create a new rc4 tarfile for HPC11.

PR #1074, opened 2022-04-27
Correct LSM_finalize routines

jvgeiger:
### Description

Remove argument "n" from definition of LSM_finalize routines.

The LSM_finalize routines do not take arguments. Also "n" clashes with
a local variable used to loop over LIS_rc%nnest.


emkemp:
Code looks good, and ifort -warn finds no problems.  I approve.

Issue #1085, opened 2022-05-05
Allow LVT to do 3-month seasonal average based on Water Year


dpsarmie:
Currently, the 3-month average option in LVT works using the meteorological seasons (DJF, MAM, JJA, SON). A new option could be added to LVT to allow for water year / calendar year triplets (JFM, AMJ, JAS, OND) to offer more flexibility to the users.

PR #1086, opened 2022-05-05
Add LVT Seasonal average for 3-month calendar/WY

dpsarmie:
This commit will add an option to LVT to allow the user to do a 3-month seasonal average via calendar/water year monthly triplets (JFM, AMJ, JAS, OND). The master config file now reflects the added option of '3 monthly WY' as a valid option in the LVT config file.

Resolves issue #1085.

### DIFF
--- a/lvt/configs/lvt.config.adoc
+++ b/lvt/configs/lvt.config.adoc
@@ -697,12 +697,13 @@ Stratification attributes file:         ./strat_attribs.txt
 
 [cols="<,<",]
 |===
-| Value     | Description
+| Value        | Description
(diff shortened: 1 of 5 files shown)


bmcandr:
Hi @dpsarmie, do you have a testcase available for this change?

dpsarmie:
Ok, I've put together a test case for you. All the files needed to run the case should be in the following directory:
/discover/nobackup/projects/fewsnet_lis/users/dpsarmie/LVT_3moWY_case

Let me know if you run into any issues.

Issue #1092, opened 2022-05-12
Add total runoff to NoahMP401


karsenau:
Add new total runoff option to NoahMP401 LSM in LIS, borrowing initial code from the AWRAL LSM implementation and expanding for different units and options.

MODEL OUTPUT table option:
qtot:         1  mm      -    0 0 0 1 0 1 # total discharge to stream

PR #1093, opened 2022-05-12
Add total runoff output field in NoahMP401

karsenau:
Added new total runoff option into NoahMP401 LSM in LIS, which
  combines surface and baseflow runoff fluxes.
 Expand this output option to include different units and
  user options.

 Add to your model output table as:

qtot: 1 mm - 0 0 0 1 0 1 # total discharge to stream

This addresses GitHub issue #1092 . 

### Description

### DIFF
--- a/lis/configs/MODEL_OUTPUT_LIST.TBL.adoc
+++ b/lis/configs/MODEL_OUTPUT_LIST.TBL.adoc
@@ -144,7 +144,7 @@ Evap:         1  kg/m2s  UP   1 0 0 1  57 10000   # Total evapotranspiration (kg
 Qs:           1  kg/m2s  OUT  1 0 0 1 235 10000   # Surface runoff (kg/m2s)
 Qrec:         0  kg/m2s  IN   1 0 0 1 163 10000   # Recharge (kg/m2s)
 Qsb:          1  kg/m2s  OUT  1 0 0 1 234 10000   # Subsurface runoff (kg/m2s)
-Qtot:         1  kg/m2s  OUT  1 0 0 1 235 10000   # Total runoff (kg/m2s)
+Qtot:         1  kg/m2s  OUT  1 0 0 1 235 10000   # Total runoff or discharge to stream (kg/m2s)
 Qsm:          0  kg/m2s  S2L  1 0 0 1  99 10000   # Snowmelt (kg/m2s)
 Qfz:          0  kg/m2s  L2S  1 0 0 1 130 10000   # Refreezing of water in the snowpack (kg/m2s)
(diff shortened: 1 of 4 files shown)


karsenau:
This new option associated with a NoahMP401 MODEL_OUTPUT_LIST.tble file has been tested with:

qtot:         1  kg/m2s  OUT  1 0 0 1 235 10000   # Total runoff (mm/s)

Please let me know if you need anything else for testing prior to merging.

dmocko:
Hi @bmcandr 

Based on the committed code, I wouldn't expect it to work.  I feel that the units of "kg/m2s" and "kg/m2" are appropriate
for the Noah-MP-4.0.1 LSM.  The units of "mm" are appropriate for the AWRAL model, which is why all three units are in
the code: lis/core/LIS_histDataMod.F90

I will let @karsenau confirm, but I believe the entry in the first comment was a typo, rather than something that needs
changing in the code.

For consistency with other runoff variable names, I would prefer that "qtot" be replaced with "Qtot" on Lines 5533
and 5535:
    https://github.com/NASA-LIS/LISF/blob/master/lis/core/LIS_histDataMod.F90#L5533-L5535

AWRAL users will need to change their MODEL OUTPUT TBL - and we should change the "qtot" entry here as well:
    https://github.com/NASA-LIS/LISF/blob/master/lis/configs/MODEL_OUTPUT_LIST.TBL.adoc

Just a suggestion, and no worries if others think we shouldn't bother.

bmcandr:
Thanks @dmocko! 

I also thought about requesting the `qtot` -> `Qtot` change, but wasn't sure whether that would affect current uses of `qtot`. Looks like the [model output table for the AWRAL testcase](https://github.com/NASA-LIS/LISF/blob/5270562e127ea4bce6ad2ab0e28c36fbad8b4207/lis/testcases/surfacemodels/land/awral.6.0.0/AWRAL600_OUTPUT_LIST.TBL#L126) would need to be updated if this change is made, for example.

karsenau:
> @karsenau and @dmocko, I made the change to capitalize `Qtot` and updated the other files accordingly. You can view the changes in my commit [here](https://github.com/NASA-LIS/LISF/pull/1093/commits/2ded466341d0c8b4c8506935e685e93240e9e3b7). Let me know if I missed anything.
> 
> Edit: should the description in MODEL_OUTPUT_LIST.TBL.adoc be "total runoff" or "total discharge to stream"?

We checked with @agetiran, and he thinks they are the same thing, just named differently.  Perhaps add it as:

"total runoff or discharge to stream"?

bmcandr:
Thanks @karsenau! Just made the change. I think we're good to merge now.

bmcandr (approved):

PR #1096, opened 2022-05-16
Expand path length in a few more places

bmcandr:
### Description

I missed a few path length variables in #1014 and fixed them here. Thanks to @karsenau for bringing this to my attention. Sorry I didn't get these in before you merged, @jvgeiger!


jvgeiger:
No problem.  Thanks for embiggening the filename lengths.

PR #1098, opened 2022-05-18
Silence err message when output dir already exists

bmcandr:
### Description

After #1042, an error message may be printed if a LIS process attempts to create an output directory that was already created by another process. This fix silences that error message by checking that `errno != EEXIST` after calling `mkdir` (suggested by @emkemp).

### Testcase

Reproducible with any LSM testcase run on multiple processes.

### DIFF
--- a/lis/core/LIS_create_subdirs.c
+++ b/lis/core/LIS_create_subdirs.c
@@ -144,19 +144,24 @@ int FTN(lis_create_subdirs) (int  *length,
 
                 /* Handle error */
                 if (mkdir_rc == -1) {
-                    /* Attempt to report error message */
-                    error_string = NULL;
-                    error_string = malloc((length_copy+1+21)*sizeof(char));
-                    if (error_string != NULL) {
-                        sprintf(error_string,
+                    /* EEXIST is returned via errno if the directory
+                    * already exists. This isn't an error because
+                    * we are attempting to create the directory.*/
+                    if (errno != EEXIST) {
+                        /* Attempt to report error message */
+                        error_string = NULL;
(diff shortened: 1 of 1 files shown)


emkemp:
It would be best to test this patch with Sujay's h2distout test case.  However, I do not have read permission for /discover/nobackup/projects/eis_freshwater/svkumar/Tests/LIS_test_distio_hymap.  Can this be moved to another location?

bmcandr:
@emkemp, I have the testcase for #1051 set up so I can run this test. Thanks for the suggestion.

emkemp:
Okay.  On my end, I have a little concern about initializing errno.  I can't tell from the mkdir(2) man page what value is assigned (if any) if mkdir runs successfully.  It may be best to set errno = 0; before the mkdir call.

bmcandr:
I'm seeing this on `mkdir(2)`:

>RETURN VALUE
>       mkdir() and mkdirat() return zero on success, or -1 if an error occurred (in which >case, errno is set appropriately).

emkemp:
Ah, okay, I wasn't reading your patched code correctly.  Yes, I expect this will work.

bmcandr:
While it should be ok, I've modified the code to match the check for `stat` earlier in the file. `errno`'s value will only be checked if `mkdir` has returned -1, thus ensuring that `errno` has been initialized.

Testing quickly and then I'll push the revised version up. Thanks again for the feedback, @emkemp.

bmcandr:
Just (force) pushed the updated changes. Compiles and tests without issue with Intel21.

PR #1102, opened 2022-05-23
Add GitHub Actions workflows to deploy docs to GitHub Pages and create PDFs on new releases

### LINKED ISSUE #1028: Use GitHub Actions to build docs into HTML and deploy to GitHub Pages
Currently, users must install an AsciiDoc Preview browser extension to view a rendered version of our LISF Users Guides and other documentation.

Instead, we can use GitHub Actions to automatically build the AsciiDoc files into HTML and deploy them via GitHub Pages for viewing at `https://nasa-lis.github.io/lisf` after any change is made. We can also automatically build PDF versions of the docs [for each release](https://github.com/marketplace/actions/upload-files-to-a-github-release).

I have set up a basic proof of concept of the GitHub Pages approach on my fork of LISF: https://bmcandr.github.io/LISF/
(the README link is broken, but the Users' Guides work)

The workflow is captured [in a YAML file stored in the `.github/workflows` directory](https://github.com/bmcandr/LISF/blob/feature/gh-pages-docs/.github/workflows/build-docs.yml).

More work is needed to refine this approach, but it is a good way to learn about GitHub Actions which we may use to automate processes in future (e.g., building container images).

----

~There is a (small?) elephant in the room: cost. [Free-level accounts can use 2000 minutes of Actions per month free of charge.](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) Once the limit is reached, no more minutes will be used unless billing is set up (i.e., no way to exceed free allocation).~

~[Building and deploying the docs takes about 1 minute](https://github.com/bmcandr/LISF/actions) so it is _extremely_ unlikely we would even get close to the 2000 minute limit in any given month.~

Edit: I misread the pricing docs. Actions are *free* for public repositories. No risk of charges at all.

bmcandr:
### Important Notes

* This should be the last PR merged into master before the 7.5 release to avoid merge conflicts or other headaches caused by filename changes in `docs/`.
* GitHub Pages must be enabled for the LISF repository for the web-based docs to work. Under _Settings_ > _Pages_, set the _Source_ branch to `gh-pages`. **Make this change before merging so the web docs workflow runs on merge.**

#### Workflows

This pull request adds two [GitHub Actions](https://github.com/features/actions) workflows to automate publishing rendered documentation:

* `build-web-docs.yaml`: when a pull request is merged into `master` AND any `*.adoc` file has been modified, this workflow converts all LISF guides to HTML and deploys them to [GitHub Pages](https://pages.github.com/) where they will be accessible at https://nasa-lis.github.io/LISF. A working example of the deployed docs based on this feature branch is available here: https://bmcandr.github.io/LISF/. This workflow can also be manually initiated from the _Actions_ tab of the repo if needed.
* `build-release-docs.yaml`: when a new release is created, this workflow converts LISF User Guides to PDFs and attaches them to the release alongside the packaged source code. See an example release with docs attached here: https://github.com/bmcandr/LISF/releases

What does it look like when these workflows run? Take a look [here](https://github.com/bmcandr/LISF/runs/6548456116?check_suite_focus=true). Past workflow runs will appear under the _Actions_ tab of the repository.

#### Other changes

In order to simplify the workflows, I also made changes to standardize the structure of some document directories:

* Where necessary, I changed the root document file for a given guide to match the name of the enclosing directory (e.g., `LIS_usersguide.adoc` --> `LIS_users_guide.adoc`. This allows us to easily build the path to the root file given the directory name.
* Moved image files for the Working with GitHub guide into an `images/` subdirectory.

References to moved/renamed files were also updated accordingly.

### Additional Notes

* GitHub Actions is free for public repositories and has no maximum quotas for minutes or storage space used ([see here](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#about-billing-for-github-actions)).
* GitHub Pages is also free, but has [some usage limits](https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages#usage-limits) that I don't expect us to exceed (there is no penalty or charge if limits are exceeded, GitHub just won't serve your page).
* I added instructions for adding a new guide to our docs to `docs/README.adoc`. Follow these instructions to ensure any new guides are picked up properly by the workflows.

----
Resolves #1028


emkemp:
I'm going to defer this pull request to @jvgeiger.

bmcandr:
I'll add that this feature is completely optional. I used our docs as an opportunity to learn how to use GitHub Actions and felt the result may be useful. Using Actions does add a little complexity to our repo that may need to be maintained, but I tried to make the workflows as simple as possible. They are designed such that they should _not_ fail as long as (1) the docs are organized/named as described in the new guidance in the README doc and (2) each document can be built successfully by `asciidoctor`. To handle point 2, I could also add a workflow to test building the docs anytime a PR is opened that includes changes to any *.adoc file and notifies the submitter if the test fails so they can fix any issues.

I'm open to any thoughts and concerns.

PR #1108, opened 2022-06-03
Fix compiling Bratseth retuning code using Cray compilers

jvgeiger:
### Description

This removes multiple uses of ESMF within two program files.  The multiple uses of the ESMF module led to a compiler error with the Cray compilers.


emkemp:
The compiles successfully on Discover w/ Intel compilers.  I approve.

Issue #371, opened 2019-10-08
Filename length limited to 100 characters 


jmroth:
It seems that the length of filenames is limited to 100 characters. This is not a problem if one uses relative paths. Absolute paths on clusters however tend to be very long, which results in LIS not finding input files. 
I think a possible fix would be to change the maximum string length for input files in the lisrc declaration, i.e. change lines like 
```Fortran    
     character*100, allocatable :: shdmaxfile(:) 
     character*100, allocatable :: shdminfile(:) 
```
to 
```Fortran     
     character*200, allocatable :: shdmaxfile(:) 
     character*200, allocatable :: shdminfile(:) 
```

Since I am not sure if this breaks anything, I will work with relative paths for now but might send a pull request in the future.

emkemp:
Thank you for opening this issue. The filename length has been on our unofficial TODO list for a while. My current thinking is to define a global integer parameter with a value of 255 (the max file length guaranteed by the POSIX standard, which many operating systems follow), and to use that parameter everywhere.  

Also, please note that the max line length in the lis.config file is 250 characters (specified by the ESMF library).  This can also effectively constrain the length of a file path if the full path is specified in the lis.config file.

bmcandr:
Should we define (or replace) our internal constant for max path length with ESMF's max path length constant (`ESMF_MAXPATHLEN`)?

----

I recently made a fix for this issue in LIS' metforcing readers (#944) by defining a path length constant ([`LIS_CONST_PATH_LEN`](https://github.com/NASA-LIS/LISF/blob/90238111aa94ba45187852ea4a9361ced238f3f7/lis/core/LIS_constantsMod.F90#L28)) of 500 characters to declare the size of strings that hold paths. 500 characters was selected by doubling the maximum line length of 250 characters listed in [ESMF's resource/config file documentation](https://earthsystemmodeling.org/docs/release/latest/ESMF_refdoc/node6.html#SECTION06090000000000000000) to provide space for the relative path to individual files that LIS tacks onto the paths read from the config. However, ESMF's documentation appears to be out of date and the max line length is actually 1024 characters as of v7.1.0:

[...]

PR #1119, opened 2022-06-10
Expand length of paths

bmcandr:
### Description

Expands the length of paths in LDT to `ESMF_MAXPATHLEN` (1024 char).

Related to #371

### Testcase

Tested with Intel 21 using the internal regression test suite.

### DIFF
--- a/ldt/ANNdata/GCOMW_AMSR2_TB/GCOMWAMSR2TB_ANNdataMod.F90
+++ b/ldt/ANNdata/GCOMW_AMSR2_TB/GCOMWAMSR2TB_ANNdataMod.F90
@@ -19,6 +19,7 @@ module GCOMWAMSR2TB_ANNdataMod
 ! !USES: 
   use ESMF
   use map_utils
+  use LDT_constantsMod, only : LDT_CONST_PATH_LEN
 
   implicit none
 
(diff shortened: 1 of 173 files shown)


emkemp:
This pull request compiles with Intel 21 w/o issue, and most of the code itself looks fine.  However, I see some apparently unrelated changes to:

ldt/metforcing/geos5fcst/read_geos5fcst.F90
ldt/metforcing/gfs/create_gfsfilename.F90
ldt/metforcing/nldas2/get_ncep_filenames.F90

Do we have test cases to double check the new filenaming logic?

bmcandr:
The changes in those files are related. The original routines built file paths using Fortran print formatting with fixed character length (e.g., `(a80)`). This wouldn't play nice with assumed length strings which are needed to avoid truncating longer paths. The original routines were also needlessly complex...

I tested the changes by writing a small Fortran program that compares the output of the new implementation against the original to ensure the paths are still being built correctly. I can attach it if needed for testing.

emkemp:
My vote (@jvgeiger can overrule) is that you upload the test program so I can look at the code.

bmcandr:
Here you go @emkemp. I had to change the file extension to `.txt` to sidestep GitHub's file filter. Just change it back to `.F90` and `gfortran` it to compile it.

The main program calls routines containing the original and new implementations for each A/B routine and compares the output. It also runs the new routines with a data directory path that is long enough to break the original routines.


[test_nldas2_ncep_fix.txt](https://github.com/NASA-LIS/LISF/files/8894835/test_nldas2_ncep_fix.txt)

emkemp:
Thanks @bmcandr!

I compiled and ran the test code after fixing a typo (subroutine name).  I satisfied that the NLDAS2 changes work.  I'm also satisfied that the GEOS and GFS logic follows the new NLDAS2 logic, so I will approve this pull request.

PR #1124, opened 2022-06-14
Add LISF Installation Guide

jvgeiger:
### Description

Add LISF Installation Guide.

### DIFF
--- a/docs/LISF_installation_guide/LISF_installation_guide.html
+++ b/docs/LISF_installation_guide/LISF_installation_guide.html
@@ -1,0 +1,2535 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<meta charset="UTF-8">
+<meta http-equiv="X-UA-Compatible" content="IE=edge">
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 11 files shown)


docs/LISF_installation_guide/LISF_installation_guide.html
```diff
@@ -0,0 +1,2535 @@
+<!DOCTYPE html>
```
bmcandr, line 1: This file should be removed as it will be built with the GitHub Actions workflow on merge to master.

bmcandr:
`asciidoctor` builds this as HTML without issue. Once the committed HTML file is removed this can be merged.

bmcandr:
Oh, and this guide should probably be linked from `docs/README.adoc`. The URL could also be inserted into the build sections of the LISF Users Guides so it is easier to get to. Something like this:

```adoc
Please first read the companion document link:https://nasa-lis.github.io/LISF/LISF_installation_guide/LISF_installation_guide.html[_LISF Installation Guide_,window=_blank].
```

(the `window=_blank` opens a new tab when the link is clicked)

jvgeiger:
Thanks!

PR #1140, opened 2022-07-06
Fix bug in handling missing IMERG files.

emkemp:
### Description

Previous code would erroneously save numerous "zero precip" observations and
pass to the Bratseth scheme if no IMERG data were read.  This has been
corrected.


emkemp:
I prefer -9999. There is logic in the code to treat accumulations less than zero as "missing", and it is possible the udef could be set to a positive number, so....

jvgeiger:
Thanks.

PR #1156, opened 2022-07-22
Replace hard-coded value

jvgeiger:
### Description

Use the generic value LIS_rc%glacierclass to check whether a tile is
a glacier type.  The hard-coded value 15 is specific to only one of our
land-cover classifications.


dmocko (approved):
Looks great - thanks!

PR #1179, opened 2022-08-04
Changes the lat/lon fields in LIS output from 2d to 1d

### LINKED ISSUE #1178: Change the lat/lon fields in the LIS output from 2d to 1d
Replace this text with a concise description of the proposed enhancement.

sujayvkumar:
With this change, the lat/lon fields will be dimensioned as:

float lat(north_south);
float lon(east_west);

Resolves #1178

### Description


sujayvkumar:
There is no specific example. Any existing LIS setup could be run with this. The only change would be in the lat/lon fields.

jvgeiger:
@emkemp Here is an LDT config file that processes LIS output:

lis/configs/557WW-7.4-FOC/NRT_GLOBAL/for_discover/ldt.config.global.noah39.cdf

emkemp:
We cannot accept this pull request as-is.  The code change requires LIS to always run with a map projection with independent latitude/longitude coordinates (like the equidistant cylindrical, "lat/lon" projection we commonly use).  But Lambert Conformal projection requires two-dimensional lat/lon fields, and accepting this pull request will break that capability (and compatibility with NU-WRF).

A possible solution is to provide the map projection type to the low-level LIS_historyMod.F90 routines, so the decision on 1d vs 2d lat and lon arrays can be made correctly.  This may require some work.

sujayvkumar:
@cmclaug2  -Sounds good overall, I suggest adding the config option, but with the assumption of the 1d option as the default - so that we can phase out the 2d option over time.

cmclaug2:
@jvgeiger @sujayvkumar @emkemp,

I updated this PR. **The lis default will now be to write 1D lat/lon fields (unless the projection being used cannot practically be written in that way).**  There is also a new lis.config option that can be used to revert to writing the lat/lon fields to 2D.  Note: Not all projections can have lat/lon fields written in 1D.  If the user asks for 1D for an incompatible projection, LIS will write a warning to the log, then write the lat/lon fields to 2D. 

One question needs to be answered before this is merged - What projections that LIS supports can be written in 1D?  More specifically, what do we want LIS to do? This can easily be changed with one line of code, but as it stands, LIS will try to write the lat/lon fields for the projections as follows:

latlon: 1D
lambert: 2D
polar: 2D
mercator: 1D
hrap: 2D
UTM: 2D

Let me know what you suggest.   Please also let me know any comments on my updates and what I can do to improve this PR.

emkemp:
The code looks good to me.  I don't think we need to drag this out any further.  I accept.

PR #1185, opened 2022-08-16
NUOPC Cap Coupling Testcase

danrosen25:
### Description

This pull request adds at test case for NUOPC coupling. It includes a small domain and LIS configuration for the Little Washita Watershed that can run on a single node. The test case one-way couples an artificial atmospheric model to LIS and then LIS to an artificial hydrology model. ATM -> LIS -> HYD. There is no need for forcing data because the artificial atmosphere prescribes meteorological forcing data.

### Testcase

The testcase is located in lis/testcases/coupling/cpl_nuopc and is ready to run on discover using lisf_7_intel_2021.4.0

### Build

1. cd lis/testcases/coupling/cpl_nuopc
2. ./configure --auto
3. ./build.sh

### Run

The build step builds the internal LIS model using the LIS compile script. Then builds the NUOPC cap and NUOPC cap library. Then finally builds the coupled NUOPC system as the lis_nuopc_test.exe executable. To run this executable enter the run directory. Modify the desired batch script template (on discover this is slurm_template.sh). Then execute the test case. A successful run outputs 3 hours of LIS history files with physically realistic data.


danrosen25:
Before accepting this pull request I'd like to review it with the LIS team.
- Is it okay to include a small testcase directly in the lis directory? The entire test case, including domain files + configuration files + source code, is ~1.1M.
- Is lis/testcases/coupling/cpl_nuopc the correct location? This can be moved to the lis/runmodes/nuopc_cpl_mode directory if needed.
- Should we automate more of the testcase, such as populating the slurm_template.sh file?

cmclaug2:
This PR compiles with intel and gnu compilers and produces 3 hours of LIS history files with physically realistic data.  I will coordinate with @jvgeiger regarding how to proceed.

PR #1207, opened 2022-09-15
Fixes to LIS log message directory lengths and reduced the NoahMP401 restart GECROS dimension

karsenau:
Added the trim function for a couple of LIS log file entries that adhere to the 256 character length.  With the long length, it can create several blank lines in the LIS log files.

Also, reduced the dimension number of 60 to 1 for the GECROS states, which were commented out in the NoahMP401 write restart file routine.


karsenau:
To @emkemp and @jvgeiger, this pull request is for LIS 7.5.

emkemp:
Code compiles with both Intel and GNU compilers w/o issue.  I approve.

PR #1210, opened 2022-09-26
Add new option to Ensemble Forecast Readers in LIS

karsenau:
Updated the ensemble forecast forcing readers in LIS with a new optional entry in the LIS config file.

### Description:

The user can now specify an initial forecast date, which supports the use of restarts in the middle of forecast run (for large domain and long-term forecast runs).  If not specified in the LIS config file, LIS resorts to default settings of the LIS start year and month.

Additional code cleanup of the "generic" and precipitation forecast readers included.

If a use case is required to complete the merge of this pull request, I can provide one.  Otherwise, the new code added does not impact any users currently using the readers and has been tested for a suite of different conditions.

### DIFF
--- a/lis/configs/lis.config.adoc
+++ b/lis/configs/lis.config.adoc
@@ -6105,11 +6105,17 @@ ensemble members desired for the user-generated ensemble forecast dataset.
 the generated ensemble-based forecast forcing files being read in.
 Default setting is GEOS5, but options include GEOS5 and CFSv2.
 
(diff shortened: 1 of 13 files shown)


karsenau:
To @emkemp and @jvgeiger, this pull request #1210 is for LIS 7.5.

emkemp:
Code builds with both the Intel and GNU compilers w/o issue.  The new lis.config settings are also documented.  I approve.

PR #1215, opened 2022-10-11
Add the LVT IMERG reader testcase

jvgeiger:
### Description

Add the LVT IMERG reader testcase


jvgeiger:
@cmclaug2 The testcase data that go with these config files are here: /css/lisdata-pub/TESTING/testcases/data/7.4/master

LVT_IMERGreader_input_v74.tar.gz
LVT_IMERGreader_output_v74-intel19.tar.gz

PR #1218, opened 2022-10-13
Add support to specify RPC library

jvgeiger:
### Description

The OS upgrade on HPC-11 changed the available low level RPC library. The system no longer has a static version to compile into the HDF4 library.  You must now specify the RPC library in the LDFLAGS variable (both when building the HDF4 library itself and when compiling LISF).

This commit adds LDT_RPC, LIS_RPC, and LVT_RPC environment variables. If these environment variables are defined, then the build scripts add "-ltirpc" to the LDFLAGS variable for HDF4.  If these environment variables are not defined, then the build scripts function as before (nothing new is added to the LDFLAGS variable for HDF4).

### DIFF
--- a/ldt/arch/Config.pl
+++ b/ldt/arch/Config.pl
@@ -586,6 +586,13 @@ else{
    $libjpeg = "-ljpeg";
 }
 
+if(defined($ENV{LDT_RPC})){
+   $librpc = "-ltirpc";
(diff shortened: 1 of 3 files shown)


emkemp:
Compiles on Discover w/o issue.  I approve.

PR #1223, opened 2022-10-25
Address errors raised by GNU compilers

jvgeiger:
### Description

The GNU Fortran compiler raised errors when compiling source code that included mpif.h (needed when compiling with gfortran while using the Intel MPI library) or used the HDF-EOS2 library.

The workaround is to add `-fallow-argument-mismatch` to the gfortran compiler flags.

Note: On some systems, this compiler option is already added by the mpif90 wrapper script.

### Testcase

I compiled on narwhal and in my virtual machine.  To test compiling on discover, compile serially.  `mpif90` on discover adds `-fallow-argument-mismatch` to the compiler options.

### DIFF
--- a/ldt/make/Makefile
+++ b/ldt/make/Makefile
@@ -124,6 +124,17 @@ version.F90:
 	@echo "Building version file $@"
 	sh build_version.sh
 
+ifeq ($(LIS_ARCH),linux_gfortran)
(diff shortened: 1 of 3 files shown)


cmclaug2:
Compilation tests performed on Discover with Intel and GNU.

PR #1226, opened 2022-10-26
Fixes for compiling with Cray compilers

jvgeiger:
### Description

This resolves compiler errors raised by the Cray compilers.


emkemp:
Code compiles successfully on Discover (Intel and GNU compilers) and HPC11 (Cray compiler).  I do notice some warning message about invalid compiler flags when the C compiler is invoked on HPC11 (apparently it is the clang compiler), but that is outside the scope of this pull request.  I approve.

PR #1231, opened 2022-10-31
Add GALWEM (17km or 0.25deg) deterministic forecast reader

yyoon4:
### Description

Add GALWEM (Global Air-Land Weather Exploitation Model) medium-range forecasts reader for GHI.
Part 1: Add GALWEM (17km or 0.25deg) deterministic forecast reader (#1230 -1) 

### Testcase

/discover/nobackup/yyoon4/lis7/Testcase/galwem

PR #1235, opened 2022-11-08
Fixed name clash introduced by PR #1231.

emkemp:
### Description

Renamed two subroutines in AGRMET code to avoid name clash in metforcing/galwem.  The name clash only occurs when AGRMET is compiled into LIS (not done by default).


emkemp:
In the interest of time, I'm going to accept my own pull request.  

BREAKING THE LAW BREAKING THE LAW
BREAKING THE LAW BREAKING THE LAW

Issue #1264, opened 2023-01-24
Need to add "micromet" option to LIS met forcing readers


karsenau:
LIS met forcing readers with terrain maps need to be updated to account for the addition of MicroMet topographic downscaling routines

@karsenau

karsenau:
Several metforcing readers have coded “ .ne. ‘none’ “ for “met_ecor” option (set in the lis.config file) when activating elevation correction to forcing fields, i.e.,:
       if( LIS_rc%met_ecor(findex).ne."none" ) then
 
The readers that currently have that as the option include:
 
Princeton forcing reader
NAM-242 forcing reader
WRFoutv2 reader
WRFAKdom reader
GFS
ERA5
ECMWF
NLDAS2

cmclaug2:
This issue was addressed in #1265

PR #1265, opened 2023-01-26
Fix/micromet metreaders

karsenau:
 Two of the LIS met forcing readers did not support the new
  MicroMet topographic downscaling option, which included:

  GDAS and MERRA2

 This MicroMet option has been included and tested into those
  two readers.

If you would like a test case for this, I would be happy to provide one.

This addresses issue #1264

### DIFF
--- a/lis/metforcing/gdas/get_gdas.F90
+++ b/lis/metforcing/gdas/get_gdas.F90
@@ -224,10 +224,13 @@ subroutine get_gdas(n, findex)
      gridDesci(10) = 95
      gridDesci(20) = 0
 
-! EMK Bug fix
-!     if(LIS_rc%met_ecor(findex).eq."lapse-rate") then 
-     if(LIS_rc%met_ecor(findex).eq."lapse-rate" .or. &
-          LIS_rc%met_ecor(findex) .eq. "lapse-rate and slope-aspect") then 
+! KRA - Previous:
+!     if(LIS_rc%met_ecor(findex).eq."lapse-rate" .or. &
+!          LIS_rc%met_ecor(findex) .eq. "lapse-rate and slope-aspect") then 
+! Including MicroMet downscaling:
+     if(LIS_rc%met_ecor(findex) == "lapse-rate" .or. &
+          LIS_rc%met_ecor(findex) == "lapse-rate and slope-aspect" .or. &
+          LIS_rc%met_ecor(findex) == "micromet" ) then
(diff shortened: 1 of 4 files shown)


cmclaug2:
Hi @karsenau,

These changes look straightforward; however, can you still provide a testcase?  Thank you!

karsenau:
@cmclaug2 - sure thing.  Since two readers are involved, would you want a test case for both, or would a test case for one of the readers suffice for the test?

cmclaug2:
@karsenau Would you mind rerunning your test?  I am producing output that is different than yours and have not been able to figure out why yet.

Thanks!

karsenau:
Thanks @cmclaug2 for reporting that. I realize why you saw some minor differences (I forgot to turn off a change that I made within NoahMP401 that was in my LIS executable). I just undid that change and reran the test experiment.  New output found here:
/discover/nobackup/karsenau/LISF_PR/LISF_MERRA2_MM/output_target

cmclaug2:
Thank you for the updates @karsenau.   Interestingly, I am still reporting minor differences in our output across most variables.  Would you mind taking another look?  I will also do a double check on my end to make sure everything is running as expected.  If I don't hear back, have a great weekend!

karsenau:
Thanks for checking again, @cmclaug2 

I'm compiling with all default configure script settings and with the following module:
lisf_7.5_intel_2021.4.0

hope that helps

karsenau:
@cmclaug2  -- Ok, please try again.  This should hopefully resolve our differences

cmclaug2:
@karsenau Thank you! I am now able to replicate the output.  Additionally, I confirm the updates compile with intel and gnu compilers. 

One final thing, can you add a note in the LIS_users_guide documenting the new entry option for `Topographic correction method (met forcing):` ?

After that, this PR is ready to be merged.

karsenau:
Thanks, @cmclaug2, for verifying the run matches now!

The entry in the LIS user's guide is already there; please see options under:
Topographic correction method (met forcing):
https://github.com/karsenau/LISF/blob/master/lis/configs/lis.config.adoc
"micromet" is there.

MERRA2 and GDAS already had support for topographic downscaling but included within their conditional statements the addition of "micromet".

Thanks again!

cmclaug2:
@karsenau Ah yes, I was looking at an outdated version of the guide.  Thank you for pointing that out.

PR #1277, opened 2023-02-14
Add CECILL_C license for Crocus8.1

jvgeiger:
### Description

Add CECILL_C license for Crocus8.1


cmclaug2:
This PR successfully adds the CECILL_C license for Crocus8.1. Since only documentation was updated, no additional testing is needed.

PR #1287, opened 2023-03-07
Fix garbage in SNODEP filename lislog message

dmocko:
This fix will correct garbage being included in SNODEP filenames as output to the lislog files.  The character string is set to an empty string before the filename is written into it.


jvgeiger:
@dmocko Would you please try testing this without the two trim commands at lines 117 and 118?  We have run into performance issues in the past with too many unnecessary trim calls.  Thanks.

dmocko:
Hi @jvgeiger 

Yes, I tested this without the two trim commands on lines 117 and 118, and it still works as expected.

As long as lines 99 and 100 are added, garbage no longer appears in the lislog message, and the files still are read in.

jvgeiger:
@dmocko Thanks.

PR #1292, opened 2023-03-16
Correct intent error

jvgeiger:
### Description

This patch fixes an intent(out) vs intent(inout) error that resulted in glaciers melting.

In SFCDIF1_GLACIER, the variable FV was declared as intent(out). Technically, this means that FV was uninitialized.  FV was initialized for the first iteration (ITER=1), but it was not initialized for subsequent calls (iterations 2 through 5).


dmocko (approved):
Yes, nice find.  Looks good to me.  I have been looking through the new re-factored 5.0 version of this code, but haven't seen this issue yet.  But if I do find it, I will report it.

dmocko:
In the re-factored code, "FrictionVelBare" (which is the re-named variable FV) is indeed listed as "inout":

https://github.com/NCAR/noahmp/blob/master/src/ResistanceBareGroundMostMod.F90

Note that Noah-MP-5.0's structure avoids having to pass variables into and out of routines.

However, the developers note that this variable is both an input and an output for this routine.

Issue #1230, opened 2022-10-31
Medium-range forecasts for GHI


yyoon4:
**Add GALWEM (Global Air-Land Weather Exploitation Model) medium-range forecasts reader for GHI**

1. GALWEM deterministic forecast 
a. GALWEM 0.25 deg dataset
   - predictions up to 6 days ahead.
   - 6-hourly cycle
 
b. GALWEM 17-km dataset
  - predictions up to 7 days ahead.
  - 6-hourly cycle 

2. GALWEM-GE (Global Ensemble)
a. 21 members (20 members plus a control run) each from NOAA/NCEP, CMC, FNMOC
b. 40 km horizontal res (half degree lat/lon)
c. predictions up to 16 days ahead.
d. 12-hourly cycle 
e. discontinued for now; will be provided with different resolutions and members

PR #1315, opened 2023-04-17
[GHI] Update code to fit new GALWEM-GE format

yyoon4:
### Description

Update GALWEM-GE (Global Air-Land Weather Exploitation Model-Global Ensemble) medium-range forecasts reader to fit the new format ([#1230](https://github.com/NASA-LIS/LISF/issues/1230)):

Spatial resolution: 40 km -> 0.5 deg 
Number of ensembles: 21 -> 18

### Testcase

6 test cases are here:
There are 3 LSMs (Noah39, NoahMP401, JULES50) with 2 routings each (HyMAP and RAPID).

Note: Currently, JULES50 is in a queue on Discover.

/discover/nobackup/projects/usaf_lis/yyoon4/Hydro/MRF/LISF_7.5_Use_Case_all/GALWEM-GE-new

### DIFF
--- a/lis/metforcing/galwem_ge/galwemge_forcingMod.F90
+++ b/lis/metforcing/galwem_ge/galwemge_forcingMod.F90
@@ -19,6 +19,7 @@ module galwemge_forcingMod
 
 ! REVISION HISTORY:
 ! 09 May 2022; Yeosang Yoon; Initial Specification
+! 05 Apr 2023; Yeosang Yoon; Update code to fit new format
 
 ! !USES:
   use LIS_constantsMod, only : LIS_CONST_PATH_LEN
@@ -125,8 +126,8 @@ contains
     enddo
 
     do n=1, LIS_rc%nnest
-       galwemge_struc(:)%nc = 640  ! galwem-ge
-       galwemge_struc(:)%nr = 480
+       galwemge_struc(:)%nc = 720  ! galwem-ge
+       galwemge_struc(:)%nr = 361
     enddo
 
     ! 8 - key met field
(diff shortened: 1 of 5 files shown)


cmclaug2:
Hi @yyoon4,

Thank you for submitting this PR and pointing to testcases.  Once JULES50 finishes running, I will begin reviewing.

yyoon4:
JULES50 with HyMAP and RAPID runs were done.

emkemp:
Apologies, but I nuked your rapid/jules output by mistake.  Can you please rerun?

emkemp:
Update on testing (normal optimization) on my end.

hymap2/jules50:  Still in queue.

hymap2/noah39:  Results match.

hymap2/noahmp401:  Results match.

rapid/jules50:  Crashed, but error message indicates low-level MPI failure (error is from UCX, a networking driver).  I'm assuming there was a problem with a compute node. Resubmitted and waiting in queue.

rapid/noah39:  Results match.

rapid/noahmp401:  Results match.

yyoon4:
I am resubmitting the job for jules50+rapid. I will let you know if it is done.

emkemp:
My rapid+jules50 rerun crashed again with the same error from UCX.  I'm investigating further.

dmocko:
In the past, I have used --ntasks-per-core instead of per-socket.

I am not sure exactly what is a "socket" in relation to a core or a node.  But if there are 2 20-core processors per node, perhaps that means there are 2 sockets on each node?

emkemp:
Hi @dmocko 

I'm assuming there are two multi-core sockets on each compute node (same as on HPC11).  However, using --ntasks-per-socket=8 doesn't help.

It turns out I had the same problem with NRT-Streamflow runs with Rapid+Jules50, and I reduced --ntasks-per-node=8.  Two test runs on Cascade Lake with this change avoid the out of memory failure.  I'm doing one more test run in the debug qos before submitting a normal job.

yyoon4:
Update: my jules50+rapid is done.

emkemp:
Hi @yyoon4 

I cannot read /discover/nobackup/projects/usaf_lis/yyoon4/Hydro/MRF/LISF_7.5_Use_Case_all/GALWEM-GE-new/output.  Please update permissions.

yyoon4:
I have changed the permission.

emkemp:
Thanks @yyoon4!

emkemp:
Okay, I compared the results between the supplied test cases and my own tests.  

The only difference detected is with the JULES50 restart file (for the HYMAP/JULES run), and that appears to be due to erroneously changing the decomposition and MPI task count in my test run.  (@yyoon4 used 384, but I used 512).  The differences are *only* in the lat and lon array values, which are *not* read back in when restarting the job (lat and lon are instead read from the LDT parameter file).  I confirmed this by using the output from the @yyoon4 supplied test case and my own test as new restart files (renamed to appear valid at 12Z 27 Mar 2023).  The history outputs of these reruns are identical, even when the MPI decomposition is changed.

The differences in lat/lon in the restart file are apparently due to a bug in LIS_gather_patch_vector_output (which gathers lat/lon data from each MPI task into a common array for the global domain).  But since this has no affect on subsequent restarts, and since this is not a bug introduced by this pull request, I believe fixing the bug can wait.

HOWEVER, the current code is not fault tolerant -- it will only accept GALWEM-GE data that starts at the same start time as the LIS run itself.  So, we will wait for additional changes to make the reader fault tolerant.

emkemp:
After further discussion, we will accept this pull request as is.  Fault tolerance will be handled in a separate future pull request.

PR #1316, opened 2023-04-17
Fix noah36 runoff

jvgeiger:
### Description

Correct a typo in noah36_lsm_ini to properly initialize fldsto.


emkemp:
Compiles w/o issue.  I accept.

PR #1321, opened 2023-04-20
Add GEOS-IT as a metforcing for LIS

### LINKED ISSUE #1320: Need to add GEOS-IT as metforcing to LIS
The GMAO's GEOS-IT product is in production, and upcoming LIS projects will use this product as metforcing.

This issue will work on adding support in metforcing in LIS for the GOES-IT product.

dmocko:
This pull request adds the GEOS-IT hourly surface products as a metforcing into LIS.  There are options available to use forcings at different heights above the ground.

Resolves #1320

### DIFF
--- a/lis/configs/lis.config.adoc
+++ b/lis/configs/lis.config.adoc
@@ -351,6 +351,7 @@ Acceptable values for the sources are:
 |"`GLDAS`"                     | GLDAS
 |"`GFS`"                       | GFS
 |"`MERRA2`"                    | MERRA2
+|"`GEOS-IT`"                   | GEOS-IT
 |"`CMAP`"                      | CMAP
 |"`TRMM 3B42RT`"               | TRMM 3B42RT
 |"`TRMM 3B42RTV7`"             | TRMM 3B42RTV7
@@ -5580,7 +5581,6 @@ GFS domain y-dimension size:      256
 GFS number of forcing variables:  10
 ....
 
-
 [[sssec_forcings_MERRA2,MERRA2]]
 ==== MERRA2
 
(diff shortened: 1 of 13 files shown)


dmocko:
A testcase for this PR is available in:

/discover/nobackup/projects/lis/tools/pull_requests/1321

cmclaug2:
@dmocko I have ran your testcase and confirmed that the new GEOS-IT metforcing reader works as expected.

I am noticing small roundoff errors in our target output on the order of less than 10^-6.  I am running with lisf_7.5_intel_2021.4.0.  If @jvgeiger is okay with these differences, I will merge this and move your testcase to our staging area to add to the test suite.

jvgeiger:
@cmclaug2, I suspect that the small roundoff errors are from the compiler flags, as in you used "-g" and @dmocko used "-O2".  For a quick test, try using the other compiler flags.

dmocko:
I copied my configure.lis file into the same directory as the testcase directory above for this PR.

I did use "-g", but it still may be different.  I am using this module file: lisf_7.5_intel_2021.4.0

cmclaug2:
@dmocko,

I am unable to manually replicate the compiler flags you used.  However, when I compile using your configure.lis file, I successfully reproduce your output.

dmocko:
@cmclaug2 - Thanks for the update.

Should I use a more standard configure.lis file, and re-build my testcase?

jvgeiger:
What are the differences between the two configure.lis files?

dmocko:
I added `"-fp-model precise"` to CFLAGS, FFLAGS77, and FFLAGS.

I also changed `"-nomixed_str_len_arg"` to `"-nomixed-str-len-arg"`.

jvgeiger:
@dmocko Would you rebuild your test case compiling with "Optimization level (-3=strict checks with warnings, -2=strict checks, -1=debug, 0,1,2,3, default=2): " -1 ?

Thanks.

cmclaug2:
I used default settings with `optimization level = -1` and noticed the flag differences David cited. 

I then tried default settings with `optimization level = -1` AND `parallelism =  0-serial`.  This added the `-fp-model precise` flag but removed the `-g` flag.

cmclaug2:
@jvgeiger and @dmocko,

I confirm that I can reproduce David's new output without problem.  I had @emkemp review my commit to avoid out of bound array errors when more than one nest is used.  This PR is ready to accept.

PR #1322, opened 2023-04-24
Update the LISF AppImage workflow for the master branch

jvgeiger:
### Description

This adds PETSc to the compile_lisf script.  PETSc support is disabled because this library has not yet been installed into the lisf_libraries image.


cmclaug2:
@jvgeiger I was able to successfully build this LISF appimage.

PR #1330, opened 2023-05-03
[GHI] Fix/lisf75 foc nrt streamflow

emkemp:
### Description

Added new FOC collection of files for NRT_Streamflow GHI configurations.

NOTE:  rapid_param files are excluded due to the large size of the files (GitHub won't accept them).


cmclaug2:
@emkemp

I noticed that the ldt.config files are stored in the lis folder.   This was done in the `./NRT_GlOBAL` folder as well.   As long as this is okay, I do not have any other comments. I see that Jim as self-assigned this PR so I will leave it here.

jvgeiger:
Hello @emkemp,

I see several empty files.  For example, lis/configs/557WW-7.5-FOC/NRT_STREAMFLOW/JULES/ps41_nml_15min/crop_params.nml

Thanks,
Jim

emkemp:
Hi @jvgeiger 

It looks like those same nml files are also zero-byte size in NRT_GLOBAL.  Should I remove them all?

dmocko:
I believe that those files still need to be present for the JULES namelists to be read, but there is no issue if they are empty.

jvgeiger:
Thanks David!

jvgeiger:
Hello @cmclaug2,

Yes it is unusual to put ldt.config files under the lis subdirectory, but we make an exception for these FOC configuration files.  It is much easier for us and the Air Force to keep all these files together than spread them out.

PR #1338, opened 2023-05-09
Feature/cdf transfering

### LINKED ISSUE #1331: Geolocation independent stratified soil moisture CDFs
Standard CDF matching needs long historical soil moisture data, which are domain-specific, forcing-specific, and resolution-specific.

New possible strategy: take advantage of other ancillary data such as precipitation data which is directly related to soil moisture. 
Method: stratified CDFs (both observation and LSM) based on precipitation climatology, and then performed the bias removal using CDF matching technique. 
Advantages: 
1- Scale the soil moisture based on precipitation regimes using a look-up table, and possibly transfer the CDFs between different domains.  
2- Initiate a regional SM DA simulation using an operational global restart file independent of the map projection and resolution of the global restart file.

We are planning to:
1- Enhanced the capability of LDT to generate geolocation-independent stratified soil moisture CDFs. 
2- Enhance the capability of LIS to assimilate geolocation-independent stratified soil moisture CDFs.

--assignee @LIS-navari

LIS-navari:
### Soil Moisture Bias Correction in Land Data Assimilation Using a Geolocation Independent Cumulative Distribution Function

SM Data assimilation needs unbiased observations. 
Standard method: CDF matching for each grid cell
Issue: long historical soil moisture data, domain specific, forcing specific, resolution specific
New possible strategy: take advantage of other ancillary data such as precipitation data which is directly related to soil moisture. 
Method: stratified CDFs (both observation and LSM) based on precipitation climatology, and then performed the bias removal using CDF matching technique. 
Advantages: 
-Scale the soil moisture based on precipitation regimes using a look-up table, and possibly transfer the CDFs between different domains.  
-Initiate a regional SM DA simulation using an operational global restart file independent of the map projection and resolution of the global restart file.

Resolves #1331

### Testcase
LDT 
/discover/nobackup/projects/usaf_lis/mnavari/nrl/conus_europe/nlads_conus/DA_proc_LSM/
config file:
ldt.config.p.climo.15bins
Output
LDTOUT.strat.p.climo.15bins

/discover/nobackup/projects/usaf_lis/mnavari/nrl/conus_europe/nlads_conus/DA_proc_SMAP
config file : 
ldt.config.p.climo.15bins
Output:
LDTOUT.strat.p.climo.15bins

LIS
Test case 1: 
/discover/nobackup/projects/usaf_lis/mnavari/nrl/conus_europe/nlads_conus/sim
config file:
lis.config.da.strat.p.climo.15bins
Output:
output.da.p.climo.15bins

Test case 2: 
/discover/nobackup/projects/usaf_lis/mnavari/nrl/conus_europe/merra2_eu/sim
config file:
lis.config.da.cdf.transferd.different.forcing
Output: 
output.da.cdf.transfer.diff.forcing


cmclaug2:
Hi @LIS-navari.  Thank you for this PR and testcases.

Here are my updates:
- I was able replicate the output of your **LDT** tests.
- I have queued your **LIS** tests. I will update when they complete. 
- I **_was_** able to compile your updates with the **intel** compiler (both LDT and LIS).
- I **_was not_** able to compile your updates with the **gnu** compiler (neither LDT nor LIS). 

Therefore, the next step to merge this PR is to resolve the issues raised by the gnu compiler.  I have attached screenshots of my errors below:

LDT 
- It appears you used some pre-compiler macros that are not supported by gnu compiler.  Since these are not located anywhere else in LISF, these should probably be removed (I am noticing this "#IF" and "#ENDIF" syntax in many of the files you have edited.  **Try switching this to a lowercase "#if" and "#endif" and see if that helps**.)
- A closer look needs to be taken at the array LDT_rc%stratification_data. I am not exactly sure what the error is here but there appears to be a mistake in how the array is either declared or indexed. 

LDT_fail



LIS

- Again, I am noticing an issue with the pre-compiler macro "#IF" and #ENDIF". Try lowercase. 
- I think xrange_file is declared as 4 dimensional but is referenced on line 1921 as 3 dimensional. 
- There is some issue with the way some variables are declared (eg. kk, nbins).


LIS_fail

LIS-navari:
Hi @cmclaug2 

I think solving the issue with the pre-compiler macro "#IF" and #ENDIF" will solve the other issues. 
Because the gun compiler ignores the upper case #IF and #ENDIF which is a way to comment out part of the code.

cmclaug2:
@LIS-navari. After your latest commit, I was able to recompile your code without error.  I am still waiting for the LIS runs to begin.  I will update when they are finished.

cmclaug2:
@LIS-navari 

I ran your LIS testcases using debug just to get a look at the output. 

The results show that I started to replicate your output for the merra2_eu test.   However, I am noticing differences in output for the nldas test.  I am looking into this and noticing that for the merra2_eu test you run with a different executable than the nldas test.  (LIS_lite_cdfTrans for merra2_eu vs. LIS_lite for nldas).   

Is there a reason for this?  It might explain the differences I am noticing.

cmclaug2:
@LIS-navari 

Thank you for this clarification.  It sounds like the LIS nldas testcase is not needed for our purposes since it is exercising that your updates work with (what will soon be) old code. 

One question about your merra2_eu testcase: 
This is very computationally expensive and the full 18-month testcase will take a while in the queue each time we run the testcase suite.  Can the full functionality of the new CDF format be expressed in a shorter time frame than you have in the current merra2_eu test? **i.e could we change this 18-month test to a 2-month test and still prove the new CDF format is being read properly?**   

Thanks, Mahdi.

LIS-navari:
Hi @cmclaug2 

Yes, you can run the DA test case for a couple of months and compare your results with mine. 
Test case 2:
/discover/nobackup/projects/usaf_lis/mnavari/nrl/conus_europe/merra2_eu/sim
config file:
lis.config.da.cdf.transferd.different.forcing
Output:
output.da.cdf.transfer.diff.forcing

cmclaug2:
@LIS-navari 

Thank you.   I have been able to replicate 2 months of your output for the merra2_eu testcase.  I now confirm your code compiles and testcases run without error. 

Before accepting, I have two notes:
1) I notice that in your testcase you have commented out your new lis.config entry `Number of bins in the soil moisture cdf:` Have you tested this new functionality yourself and confirmed that it works?
2) I have performed a code review and have not noticed any red flags.  However, I would like to ask @emkemp or @jvgeiger to take a second look and confirm this is ready to be merged.

LIS-navari:
Hi @cmclaug2 

We do not need this config entry. 
Number of bins in the soil moisture cdf:

emkemp:
Hi @LIS-navari 

I dusted off this pull request, and I found a serious bug in read_Precip_climo.F90.  If the path to the precip climo file is wrong, the subroutine will still try to copy data from the precip_climo array, which in this case will be unallocated.  The error handling needs to improve, otherwise LDT will just crash.

LIS-navari:
Hi @emkemp 
I have updated the code. Please let me know if I need to further modify the code.

LIS-navari:
Hi @emkemp 
I have updated the code. Please let me know if I need to further modify the code.

cmclaug2:
@emkemp @LIS-navari,

After Eric's comments and Mahdi's code updates I confirm:

- This PR compiles with intel and gnu compilers with strict checks
- I can replicate both LDT tests
- I can replicate the LIS merra2 test
- I cannot replicate the nldas test using code compiled from the Main branch as of 8/30/2023.  (Though I think this is okay. I have not tracked down why, but it makes sense that there may have been code updates since Madhi ran his test in 2022 that change the output.)  Sanity check - I was able to replicate Mahdi's output using his 2022 compiled version of LIS from the Main branch. 
- I confirm that Mahdi's bug fix in `read_Precip_climo.F90` allows LIS to fail informatively if the file path is wrong.

emkemp:
I tested this code with the Intel compilers (w/ and w/o debugging), and exactly reproduced results.

I then tested with the GNU compilers.  I consider the LIS results to be successful (output from both compilers is very, very similar, with only a handful of grid points with differences > 0.1% in Evap_tavg, Qs_tavg, SoilMoist_tavg, SoilMoist_inst, and these were with very small values of the respective variables).

I cannot run LDT with GNU Fortran -- but, this is due to a dimension error in reading monthly albedo data, in code untouched by these CDF changes.  (There are also compile time problems found with the GLASSlai and NASA_AMSRE_sm readers when I compile with debugging.  Again, unrelated to the CDF code changes.)  I don't think it is fair to penalize this pull request for these problems, but instead we should open separate issue tickets to resolve.

@jvgeiger Any objections to me accepting this pull request?

emkemp:
Hold on, this LDT runtime problem with gfortran may be due to a bad ldt.config setting.  Let me investigate further.

emkemp:
Okay, I tracked the LDT runtime error to setting the Albedo spatial transform to "none".  When I change it to a more appropriate "bilinear" the tests run.  And I get identical results compared to the Intel runs.

I'm satisfied with this code.  I will meet with @jvgeiger on Thursday to discuss handing over the test cases before merging into master.

emkemp:
I (finally) migrated the test cases to the appropriate directory.  So, I will now approve the pull request. Huzzah!

PR #1340, opened 2023-05-16
Updated script to output GeoTIFF files with 557 WW filename convention.

emkemp:
### Description

GeoTIFF files produced by make_sm_geotiff.py now follow 557 WW filename convention.


cmclaug2:
@emkemp 

These edits to the file naming convention make sense to me.  I accept.

PR #1341, opened 2023-05-17
Added missing table files for soil moisture anomaly calculation.

emkemp:
### Description

Added several missing lookup table files for LVT, for use in calculating soil moisture anomalies.

NOTE:  The METRIC.TBL.anomaly.* files are not identical because of an early mistake on Narwhal in starting the climo calculations.  Migration to a common METRIC.TBL file will require complete recalculation of the climatologies, which will be very time consuming.


cmclaug2:
@emkemp 

This PR adds missing lookup tables for LVT and includes the proper documentation within the files.  I accept.

PR #1342, opened 2023-05-17
Restored deterministic GALWEM MR lis.config files.

emkemp:
### Description

Restored older lis.config files for deterministic GALWEM MR runs.

Note:  Filenames were changed slightly to emphasize these are for MR.


cmclaug2:
@emkemp,

I notice the new set 6 lis.config files for GAWLEM medium range runs. Looks good to me.

PR #1344, opened 2023-05-23
Address error raised by Cray compilers (CPE 23.02 / CCE 15.0.1)

jvgeiger:
### Description

The Cray C compiler raised this error:

incompatible pointer to integer conversion passing 'void *' to parameter of type 'int'

In C (Cray, Intel, and GNU compilers), NULL is defined to be '((void *)0)'. In create_filelist.c, we were using NULL to represent 0.


cmclaug2:
I have verified that this fix compiles with gnu and intel compilers.

PR #1345, opened 2023-05-23
Add non-zero exit status to LISF endrun routines

jvgeiger:
### Description

Add non-zero exit status to LISF endrun routines.


cmclaug2:
ldt, lis, and lvt all compile with intel and gnu compilers.

PR #1347, opened 2023-06-02
[GHI] Updated 557WW streamflow convention.

emkemp:
### Description

Updated 557WW streamflow convention.

Based on approved update from 557WW and LCMC.


cmclaug2:
This is a straightforward fix to streamflow file naming convention.  I accept.

PR #1348, opened 2023-06-02
[GHI] Updated SM GeoTIFF filenames after consultation with 557WW and LCMC.

emkemp:
### Description

Updated soil moisture GeoTIFF filenames after consultation with 557WW and LCMC.


cmclaug2:
This pull request appears to successfully update the file naming convention for SM GeoTIFF output.   I will accept this pull request because these edits do not affect other parts of the code.  If the file naming convention needs further edits, another pull request can be submitted.

Issue #1364, opened 2023-06-26
bug fix in S2S scripts that had made 6-hourly precip values constants


smahanam:
A bug was found in s2s sripts. This bug fix should be merged to support/lisf-557ww-7.5.

PR #1365, opened 2023-06-26
fixed a bug that had made 6-hourly precip values constants

smahanam:
This is related to [Issue 1364](https://github.com/NASA-LIS/LISF/issues/1364)

### DIFF
--- a/lis/utils/usaf/s2s/s2s_modules/bcsd_fcst/bcsd_library/temporal_disaggregation_6hourly_module.py
+++ b/lis/utils/usaf/s2s/s2s_modules/bcsd_fcst/bcsd_library/temporal_disaggregation_6hourly_module.py
@@ -33,7 +33,7 @@ def scale_forcings (mon_bc_value, mon_raw_value, input_raw_data, bc_var = None):
     output_bc_data = np.ones(len(input_raw_data))*-999
 
     if bc_var == 'PRCP':
-        if mon_raw_value <= 1.e-4:
+        if mon_raw_value == 0.:
             correction_factor = mon_bc_value
             ## HACK## for when input monthly value is 0
             output_bc_data[:] = correction_factor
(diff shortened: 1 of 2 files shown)


cmclaug2:
Hi @smahanam,

Before accepting, I just want to suggest something - 

I imagine the line of code `if mon_raw_value <= 1.e-4:` might have been used to account for small rounding errors when calculating `mon_raw_value`.  (i.e, people were expecting a value of 0 for `mon_raw_value`, but they were actually finding a value that was functionally 0, but defined as something slightly bigger or smaller due to rounding errors) 

Might it be best to keep that fault tolerance of `1.e-4`, but change the line of code so it reads something like `if abs(mon_raw_value) < 1.e-4:`?  This would account for rounding errors on the positive and negative side of 0. 

This is just a suggestion.  I am not aware of how this code is used or the specific bug being addressed.  If you would like to keep you bug fix as is, just let me know and I will accept.

emkemp:
@smahanam @cmclaug2 I will go ahead and merge this as is.  The old 1.e-4 threshold lead to weird results with monthly bias corrected totals, which were discovered last week.  The results were bad enough to abort showing S2S results in a monthly newsletter.

The new code should still prevent division by zero, and more importantly was tested.  And we need to ship.  Further tweaks can be considered later.

smahanam:
Hi @cmclaug2, 
mon_raw_value in the script is precipitation in kg/m2/s (or mm/s).  
1.e-4 turned out to be a bad threshold because  precipitation in kg/m2/s is always less than 1.e-4. This 1.e-4 threshold resulted in a fix value for the 6-hourly time series.

You are right about the rounding error.

smahanam:
@emkemp -- since this work is time sensitive, yes, for now, please go ahead with  == 0.

PR #1367, opened 2023-06-27
Replaced call to LDT_endrun to avoid erroneous error code 1.

emkemp:
### Description

Replaced call to LDT_endrun at end of USAFSI_run to avoid erroneous error code 1.

LDT_endrun now returns error code 1, indicating abnormal termination. Thus, a normal termination should let the executable progress to the end of the program.

This bug was reported by TADS.


cmclaug2:
Although probably not needed for this simple fix, I confirm this fix compiles with intel and gnu compilers.

PR #1368, opened 2023-06-29
Disable AWRAL in user_557ww.cfg and user_ghis2s_557ww.cfg

jvgeiger:
### Description

SonarQube reports an "Out of bound memory access" finding for AWRAL. A review of the code showed that that was a false positive finding. However, to remove this finding from the SonarQube report, I am disabling compiling AWRAL in LIS for the USAF configurations.  USAF does not run AWRAL in operations.


jvgeiger:
Thanks.  I am preparing a delivery for TADS now.

PR #1369, opened 2023-06-30
Disabled AWRAL.6.0.0 LSM from 557WW builds.

emkemp:
### Description

This corrects an oversight in disabling AWRAL compilation to address SonarQube "Out of bound memory access" finding on AWRAL.  Prior edits to 557WW user build config files only disabled the AWRAL metforcing reader, not the LSM.


emkemp:
Due to Air Force time constraints, I will accept my own pull request.

*breaking the law, breaking the law*
*breaking the law, breaking the law*

PR #1393, opened 2023-07-31
Updated IMERG version to V06D in FOC files

emkemp:
### Description

Updated IMERG version to V06D in FOC files. V06D went live on 2 Jul 2023.

These changes were requested by the TADS contractor.


cmclaug2:
No complaints from me.

PR #1395, opened 2023-08-07
Modify IMERG reader to ingest V07A Final products

### LINKED ISSUE #1394: Update IMERG Reader for v07
The IMERG metforcing reader will be updated to ingest V07 files. The reprocessing of the V07 Final products began in July 2023, and the Early and Late runs are expected in Fall 2023. 

https://gpm.nasa.gov/sites/default/files/2023-07/IMERG_V07_ReleaseNotes_final_230713.pdf

jerlingi:
### Description

The IMERG reader was modified to read the newly released V07 products. Only Final IMERG V07A products are available at this time (August 2023), but Early and Late products are anticipated in Fall 2023. In V07 and subsequent versions, variable names have changed. Please see https://gpm.nasa.gov/sites/default/files/2023-07/IMERG_V07_ReleaseNotes_final_230713.pdf for more documentation. 

Resolves #1394 

### Testcase

A testcase is available at /discover/nobackup/projects/lais/sharespace/imergv07_forcing_testcase

### DIFF
--- a/lis/configs/lis.config.adoc
+++ b/lis/configs/lis.config.adoc
@@ -5973,8 +5973,18 @@ default to the "`final`" product of GPM IMERG.
 
 `IMERG version:` specifies the version of the GPM IMERG
 precipitation forcing files. If no version is specified, 
-the reader will default to the latest version of GPM IMERG.
-(Currently V06B).
+the reader will default to "`V06B`".
+
+The supported versions of GPM IMERG are:
+
+|====
+|Version  | Product
+
+|V06B     | early, late, final
+|V06C     | early, late, final
+|V06D     | early, late, final
+|V07A     | preliminary support for final
+|====
 
 .Example _lis.config_ entry
 ....
--- a/lis/metforcing/imerg/read_imerg.F90
+++ b/lis/metforcing/imerg/read_imerg.F90
@@ -22,7 +22,7 @@
 ! !INTERFACE:
 subroutine read_imerg (n, kk, name_imerg, findex, order, ferror_imerg )
 ! !USES:
-  use LIS_coreMod,only : LIS_rc, LIS_domain, LIS_masterproc
+  use LIS_coreMod,only : LIS_rc, LIS_domain
   use LIS_logMod, only : LIS_logunit, LIS_getNextUnitNumber, &
                          LIS_releaseUnitNumber
   use LIS_metforcingMod,only : LIS_forc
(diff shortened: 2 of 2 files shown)


cmclaug2:
@jerlingi Thank you for the packaged testcases.  I have been able to:

- compile your code with the intel compiler with strict checks.  
- compile your code with the gnu compiler with strict checks. 
- Read in IMERG V07A and replicate your testcase output.


Before merging, I would like to point @jvgeiger's attention to a few things for approval. 

In  **`read_imerg.F90`**:
- Line 150 and 156.  `len` is defined as a specific length.  Is that fine or is `len=*` preferred? 

In **`readcrd_imerg.F90`**
-  Line 57 and 59.  LIS defaults to reading IMERG V06B.  Should this be changed to V07A, or left as is?  In either case, I will add an additional note in the documentation before merging. 

All else looks good to me.

jvgeiger:
Thanks @cmclaug2.  I will look at that.

Also, @emkemp should look at this.  He is also reviewing IMERG V7 for our USAF project.

jerlingi:
I agree that @emkemp should have a look too. In my opinion, we shouldn't change the default to V07A until the Early and Late versions are released (~Fall 2023) because it will fail unless Final is selected.

jvgeiger:
@cmclaug2 When you pass a string into a Fortran routine, you may declare the length of the string as (len=*).  See line 148.  Fortran knows the length.  Here, the string filename was first declared in the calling routine at line 74 with a specific length (len=LIS_CONST_PATH_LEN).  So, in the called routine, you may use (len=*).

For local variables, you must specify the length.  dsetname and vlname are local to read_imerghdf. So, at lines 150 and 155, a specific length is needed.

emkemp:
I reviewed the changes by @jerlingi.  I think the code can be improved slightly by using a formatted read '(I)' instead of the unformatted *. 

However, I have more fundamental concerns about the subroutine (not @jerlingi 's fault).  Presently a return statement will be executed if any non-zero error code is returned by HDF5, *without* calling any HDF5 "close" subroutines to clean up under the hood.  I don't think that's acceptable -- h5close_f should always be called if there was a successful h5open_f call; h5fclose_f should be called if there was a successful h5fopen_f (to close the file!); and h5dclose_f should be called if there is a successful h5dopen_f call.  

I think a subsequent pull request is in order to fix these issues.

jvgeiger:
Good catch - I agree.  One more suggestion: all processes should log status messages, not just LIS_masterproc. (Again, not @jerlingi 's fault.)

jvgeiger:
I made the `'(I)'` the LIS_masterproc changes.  I will look into the various `h5?close_f` calls.

emkemp:
Thanks.  I started setting up a work directory to test changes, but I've now been sucked into emergency Air Force S2S work.   I think that has the higher priority. :-(

emkemp:
I had some time today so I made and tested some code improvements.  Please let me know what you think.

jvgeiger:
@emkemp Your commit looks good.  I added one commit on top of yours.

jvgeiger:
What versions of IMERG do we support?

V06B: early, late, final
V07A: final

emkemp:
Latest code compiles, and documentation looks fine.  I accept.

PR #1398, opened 2023-08-13
Fix for S2S-Post accumulated variables

karsenau:
Committing update to S2S-Post monthly script for accumulated variables (e.g., total precip).

Also, include minor fixes to BCSD python script headers (e.g., "NMME" to "CFSv2).


karsenau:
@jvgeiger and @emkemp:

This is the latest commit for the LIS 7.5.11 557 WW PR.

PR #1400, opened 2023-08-21
Fix GNU Fortran compiler error

jvgeiger:
### Description

Fix compiler error raised by the GNU gfortran compiler.

### DIFF
--- a/lis/metforcing/imerg/read_imerg.F90
+++ b/lis/metforcing/imerg/read_imerg.F90
@@ -165,7 +165,7 @@ subroutine read_imerghdf(n, filename, xsize, ysize, precipout, istatus)
 
   ! Variable names changed in IMERG V07
   vlname = trim(imerg_struc(n)%imergver)
-  read( vlname(3:3), '(I4)') vnum
+  read( vlname(2:3), '(I2)') vnum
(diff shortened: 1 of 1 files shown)


emkemp:
Why is the format set to I4?  Only a single digit is being read in.

(And as I think about it now, it should be two digits (V07A), which requires a change to the substring parsing.)

jvgeiger:
We are reading string into an integer.  As I understand it, the statement `read(vlname, '(*)') vnum` the format `'(*)'` specifies the default size of the integer.  We switched to using `'(I)'`, but GNU does not like that.  I changed it to `'(I4)'` because the target value is an integer*4.

I lost my Fortran reference book, and I cannot find the online source for my "understanding".  I will defer to your reference book.

emkemp:
Yeah, the language in my Modern Fortran book is not as clear as I would like.  But I have an older Fortran 90 book that describes "Iw" (where w is an unsigned int) as "Read the next w characters as an integer".  So I think it should match the number of characters being provided.

I compiled a proposed change, and the test job is stuck in the queue.  Hopefully it will run overnight and I can push.

emkemp:
This code runs w/o issue.  Let me know what you think.

Issue #1406, opened 2023-09-05
ISRIC soil texture assigns wrong index value for water in LDT


dmocko:
When choosing ISRIC soil textures in LDT, an incorrect index value will be assigned to water points.

The LDT ISRIC reader re-maps the soil textures from the source ISRIC datasets onto the STATSGO soil texture classification.

However, ISRIC assigns water points an index value of 13, while this value should be 14 for the STATSGO soil texture classification.

![ISRIC_water-points-issue_1](https://github.com/NASA-LIS/LISF/assets/38500701/136438c6-8540-4211-8dd6-6b049d80c614)
![ISRIC_water-points-issue_2](https://github.com/NASA-LIS/LISF/assets/38500701/17e87b66-72b5-4f0d-b7a6-9f2e4024c722)
![ISRIC_water-points-issue_3](https://github.com/NASA-LIS/LISF/assets/38500701/33e730d8-9db9-41e2-a5ab-38dd55b509c3)

Although this is a bug, it shouldn't greatly affect past simulations that used the ISRIC soil textures, as the affected points are very likely to be "water" points in the landcover classifications as well.  Still, users should check to ensure that this is true for their own individual past domains.

This bug revealed itself when running the MMF groundwater scheme in Noah-MP, which relies on soil textures even for water points.

Many thanks to @TimLahmers for identifying this issue.

TimLahmers:
Thanks for submitting this @dmocko Please let us know when the PR to fix the bug is added. I will update my MMF branches and retest the code with this new fix when it is ready. That should be the final test needed before I submit the MMF PR.

cmclaug2:
Resolved by #1407

PR #1407, opened 2023-09-07
Fix ISRIC soil texture water point index value in LDT

dmocko:
This pull request fixes the water point index values when using the ISRIC soil texture maps in LDT.  LDT re-maps the ISRIC soil texture map onto the STATSGO soil classification index.  However, the water points were incorrectly assigned a value of 13, when it should be 14 in the STATSGO soil texture classification.

Resolves: #1406

### DIFF
--- a/ldt/core/LDT_paramOptCheckMod.F90
+++ b/ldt/core/LDT_paramOptCheckMod.F90
@@ -483,8 +483,8 @@ contains
           endif
 
         case( "ISRIC")
-          if (nt.ne.13) then 
-            write(LDT_logunit,*) "Param_Check: ISRIC has 13 types (includes water) for tiling."
+          if (nt.ne.16) then
+            write(LDT_logunit,*) "Param_Check: ISRIC has 16 types (includes water) for tiling."
             write(LDT_logunit,*) "             The current value is: ",nt
             write(LDT_logunit,*) " Stopping ..."
             call LDT_endrun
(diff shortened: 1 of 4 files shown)


dmocko:
Note that the ISRIC sections of the code were changed to be more similar to the STATSGO sections of the code.

In ldt/core/LDT_paramOptCheckMod.F90 and in params/soils/set_ISRIC_texture_attribs.F90, the number of soil types is set to 16 (which is also done for the STATSGO soil texture classification).

In ldt/core/LDT_soilsMod.F90, the "NUMBER_SOILTYPES" global attribute is set to 19 (and not 16) - again following what is done for the STATSGO soil texture classification.

The reason for this difference (16 vs. 19) is that the STATSGO table in Noah/Noah-MP's [SOILPARM.TBL](https://github.com/NASA-LIS/LISF/blob/master/lis/configs/557WW-7.5-FOC/NRT_GLOBAL/noahmp401_parms/SOILPARM.TBL) file has 19 total number of soil types.

However, the STATSGO Native raw soil texture map only has the first 16 of these 19 soil types.

An argument could certainly be made that the "NUMBER_SOILTYPES" global attribute could be changed so it is set to 16 for both STATSGO and ISRIC.  For now, I made ISRIC be set to 19, to be consistent with STATSGO.

cmclaug2:
@dmocko I suppose you had a chance to chat with Jim about this.  Thanks for the explanation of why   16 AND 19 were used as the number of soil types for ISRIC.  Fair enough to me; I am ready to merge if @jvgeiger agrees.

PR #1414, opened 2023-09-18
Rework LAPACK and PETSC handling by the configure scripts

jvgeiger:
### Description

These changes allow LIS/RAPID to be compiled with the GNU compilers into an AppImage.

Note that this is disabled by default in the GitHub Actions scripts for the AppImage.

### DIFF
--- a/.github/scripts/appimage/compile_lisf
+++ b/.github/scripts/appimage/compile_lisf
@@ -52,12 +52,11 @@ printf '%s\n' "0" "2" "2" "2" "4" "1" "1" "1" "1" "1" "1" "1" "1" "1" | ./config
 # Use MINPACK? (1-yes, 0-no, default=0): 0
 # Use LIS-CRTM? (1-yes, 0-no, default=0): 0
 # Use LIS-CMEM? (1-yes, 0-no, default=0): 0
-# Use LIS-LAPACK? (1-yes, 0-no, default=0): 0
-# Use LIS-MKL-LAPACK? (1-yes, 0-no, default=0): 0
+# Use LIS-LAPACK? (0-no, 1-mkl, 2-lapack/blas, 3-lapack/refblas, default=0): 0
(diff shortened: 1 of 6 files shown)


emkemp:
Code changes look okay.  I did a couple of test compilations with GHI compiler settings and found no problems.  Do we have an environment module with libraries for the LAPACK options?

emkemp:
Hi @jvgeiger:

LIS compiles when using the LAPACK/REFBLAS option, or when using the "none" option.  But, it fails to build for MKL or LAPACK/BLAS options (cannot find -lmkl or -lblas).

jvgeiger:
Hello @emkemp,

Sorry, I did not explain this well.  The LAPACK prompt and PETSc prompt work together.  The following pertains to working on discover.

When using the Intel compilers, the PETSc library was compiled using Intel MKL.  When using the GNU compilers, the PETSc library was compiled using a "lapack/refblas" version of LAPACK.

To compile LIS with the PETSc library enabled for LIS-RAPID support

with Intel compilers:
```
Use LIS-LAPACK? (0-no, 1-mkl, 2-lapack/blas, 3-lapack/refblas, default=0): 0
Use PETSc? (1-yes, 0-no, default=0): 1
```
with GNU compilers:
```
Use LIS-LAPACK? (0-no, 1-mkl, 2-lapack/blas, 3-lapack/refblas, default=0): 3
Use PETSc? (1-yes, 0-no, default=0): 1
```
For the case with Intel compilers, one may select `1-mkl` as well.  In either case, the linker will find the MKL library.

So, the behaviour that you saw is expected.

emkemp:
Hi @jvgeiger 

Thank you for clarifying.  I tested the GNU module with LIS-LAPACK 3 and PETSc 1, and the standard Intel module with LIS-LAPACK 0 or 1, PETSc 1, and code compiled normally.

I will accept this pull request.

PR #1435, opened 2023-10-23
Remove user prompt and file checker re-added

karsenau:
- Commented out user prompt for downloading CFSv2 files on HPC11;
 - Readded the CFSv2 file checker (no prompt included);
 - Minor updates to script messages and comments.

 Target for LISV7.5.13 patch.


karsenau:
Updated code and tested on both Discover and HPC11. Passed different checks (with files present and files not present).

PR #1452, opened 2023-11-15
add distributed binary option for LIS_RST and LIS_DAPERT files

mlwrzesien:
### Description

Adds the ability to write LIS_RST and LIS_DAPERT files as distributed binary. This will help with memory intensive simulations, particularly with reading/writing the LIS_DAPERT files.

### Testcase
There's an example testcase available here:
/discover/nobackup/projects/eis_freshwater/mwrzesie/TestCase_distRST

This test demonstrates writing distributed binary for both LIS_RST and LIS_DAPERT files with a assimilation example over the Tuolumne watershed.

### DIFF
--- a/lis/configs/lis.config.adoc
+++ b/lis/configs/lis.config.adoc
@@ -7993,7 +7993,7 @@ of how to specify a time interval.
 `Noah-MP.4.0.1 restart file:` specifies the Noah-MP-4.0.1 LSM restart file.
  
 `Noah-MP.4.0.1 restart file format:` specifies the Noah-MP-4.0.1 restart
-file format (default = netcdf).
+file format.  Acceptable values are: "`binary`", "`netcdf`", or "`distributed binary`".
  
 `Noah-MP.4.0.1 soil parameter table:` specifies the filename of the
 Noah-MP-4.0.1 soil parameter table.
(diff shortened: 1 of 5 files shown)


mlwrzesien:
Adding a comment here following the discussion during the LIS tag up - once this is brought into the master branch, we should make a note that once the processor layout is chosen, you cannot change it on a model restart. LIS will expect the same processor layout and count for the distributed files.

karsenau:
Hi @mlwrzesien:

You can include this information, new lis config option and note (e.g., about needing to maintain the same processor layout) in the lis.config.adoc file.  Thanks!

mlwrzesien:
Right now I have wformat hardcoded in lis/dataassim/perturb/gmaopert/gmaopert_Mod.F90 because there isn't a lis.config option for specifying format for the DA restart files.

I also include an edit to lis/surfacemodels/land/noahmp.4.0.1/NoahMP401_readcrd.F90 because when I tried to fix the hardcoding in NoahMP401_writerst.F90, I found that the existing NoahMP401_readcrd.F90 actually hardcodes NOAHMP401_struc(n)%rformat to netcdf, so it was preventing the distributed binary option from working.

PR #1469, opened 2023-12-14
Fix LIS Bratseth autotuning

emkemp:
### Description

Modified Bratseth code to include obs flagged by backQC and superstatQC in the OBA files.  This breaks the feedback loop when cyclically autotuning the Bratseth error covariance settings, where obs were progressively being rejected for having large differences from the background or neighbors, leading to smaller error variances, leading to more rejections, etc.

Some bug fixes and tweaks are also included in the autotuning code.


jvgeiger (approved):

jvgeiger:
Merge is now pending further discussion with Sujay and Jerry.

emkemp:
Do not merge yet.  More evaluation required.

emkemp:
Code is ready to merge.  We will postpone the decision on whether to reactivate in Ops.

PR #1478, opened 2024-01-11
Fault tolerant version of GALWEM reader

emkemp:
### Description

Significant reorg of LIS GALWEM reader code to add fault tolerance.

This includes rolling back to the previous (started 12-hr prior) GALWEM run if the current GALWEM run does not have at least 2 bookends to temporally interpolate atmospheric forcing for LIS. Also, LIS will gracefully terminate early if not enough GALWEM files are found to complete the requested LIS run (e.g., 228 hours instead of 240 hours).

This is a DRAFT pending requirements review.

### DIFF
--- a/lis/core/LIS_coreMod.F90
+++ b/lis/core/LIS_coreMod.F90
@@ -462,10 +462,14 @@ contains
 ! \label{LIS_endofrun}
 !
 ! !INTERFACE:
-  function LIS_endofrun() result(finish)
+  function LIS_endofrun(override_end_time) result(finish)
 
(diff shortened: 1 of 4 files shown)


emkemp:
NOTE:  This code was tested with both Intel and GNU compilers.

emkemp:
Please hold on accepting this.  I need to change the LIS termination method if GALWEM cannot be found.

emkemp:
Okay, I improved the exit behavior when LIS has to terminate early.  I believe this is ready to merge.

PR #1486, opened 2024-01-26
Update codes for MOGREPS-G precipitation bias-correction

yyoon4:
### Description

**Precipitation bias correction for MOGREPS-G**

- Ref: Reichle and Koster, 2004, Bias reduction in short records of satellite soil moisture, GRL.
- Precipitation monthly climatology 
   : Generated from 15-year NAFPA precipitation (2008 – 2022)
   : Sample size: 465 (e.g., Jul.) for each grid
- MOGREPS-G monthly precipitation
   : Apply bias correction method to every lead time (8-days)
   : Sample size for every lead time: about 60 for each grid
   : To control statistical noise in the cdf estimate, 3x3 moving window is used 

Although the technique is methodologically corrected, but due to limited datasets, it produces under/over-estimated results. Therefore, for now, we do **not recommend** using this mode.   

### Testcase

/discover/nobackup/projects/usaf_lis/yyoon4/Hydro/MRF/MOGREPS-G/testcase/lis/lis.config.global.noah39_bc


emkemp:
I can reproduce the Intel results provided by @yyoon4, and the GNU results look very similar.  Runs with strict checks also look very similar.  I approve.

@jvgeiger My test runs are in /discover/nobackup/projects/usaf_lis/emkemp/pr/1486/work.

PR #1494, opened 2024-02-21
Add GDAL and FortranGIS to modulefile for narwhal

jvgeiger:
### Description

I was able to install GDAL (and thus FortranGIS) on narwhal.

This commit completes this modulefile.


emkemp:
LDT and LVT link on Narwhal with new library paths w/o issue.  I accept.

PR #1504, opened 2024-03-12
Several updates to IMERG reader in LVT.

emkemp:
### Description

* Fixed unit conversion to kg/m2.
* Added support for V07x.
* Now uses previous IMERG 30-min file, so LVT can get the 30-min accumulation ending at the LVT clock time (matches other readers)


emkemp:
Hold this pull request.  I discovered the existing reader won't work if IMERG is specified as the second dataset in LVT data comparision.

Issue #839, opened 2021-07-16
Need to add GLASS LAI config entries to LIS Users' Guide


dmocko:
The GLASS LAI DA config entries are not currently in the LIS Users' Guide, and should be added.

PR #1571, opened 2024-07-10
Add GLASS LAI and GLASS Albedo configs to the LIS Users' Guide.

dmocko:
This adds the GLASS LAI and GLASS Albedo configs to the documentation and the LIS Users' Guide.

Resolves: #839


jvgeiger:
Thanks David!

PR #1580, opened 2024-07-17
Add ESPC-D support

emkemp:
### Description

Add option to use ESPC-D output instead of GOFS for SST and sea ice.


emkemp:
This is a work in progress, not ready yet to merge.

PR #1597, opened 2024-09-04
Added inverse exponential correlation function option.

emkemp:
This has closer fits to data on Air Force global domain.  However, it should not be immediately used in Ops since the spin-up data we will provide used the Gaussian function.

In addition, the inverse exponential function has a longer tail, increasing the radius of influence around a particular location. This leads to more calculations and a slower analysis.  I will see if I can speed this up by replacing divisions with multiplications by an inverse variable.


emkemp:
I believe this is ready to review/merge.

PR #1599, opened 2024-09-10
Improved end-of-file handling in LIS Bratseth scheme

emkemp:
### Description

Added graceful handing of unexpected End-Of-File in sfcobs and preobs files used by LIS Bratseth scheme.  When encountered, LIS will stop trying to read the file in question, put a warning in the lislog file, and write an alert file.

Also modified code to skip unnecessary repeat of LIS precip analysis 15 minutes after start of LIS run.  Not only does this save time, it avoids redundant warnings and alerts for EOF cases.


emkemp:
Actually, hold on.  I need to check LIS behavior when started at 0600Z or 1800Z.

emkemp:
I believe this is ready.

PR #1610, opened 2024-09-25
Increase length of lis_config_file to LIS_CONST_PATH_LEN

jvgeiger:
### Description

Increase the length of lis_config_file to LIS_CONST_PATH_LEN.


emkemp:
Code looks good and compiles w/o issue.  I accept.

PR #1612, opened 2024-09-27
Correct interpolation in GEFS reader

jvgeiger:
### Description

This resolves an issue where the GEFS reader could misinterpolate both the swdown and the lwdown fields using the budget_bilinear/conservative method.

This also fixes a bug in the temporal interpolation of swdown by initializing zdoy right before calling zterp.

### Reforecast2 type
plot_swdown1 8 before
plot_swdown1 8 after

### Operational type
plot_swdown_f_tavg 7 before
plot_swdown_f_tavg 7 after


karsenau:
Hi @jvgeiger:

Thank you for checking these fields and making these corrections.   I have reviewed and agree with your updates made.

jvgeiger:
Thanks @karsenau.

jvgeiger:
I placed a copy here: /discover/nobackup/jvgeiger/for/eric/gefsreforecast2/

emkemp:
Thanks @jvgeiger 

Code compiles and runs w/o issue, and I reproduced your results.  Running with strict checks revealed no errors.  Since @karsenau concurs, I will accept the PR.

PR #1623, opened 2024-10-17
 Revised alert and abort message files for USAFSI

jvgeiger:
### Description

Porting PR #1622 into support/lisf-557ww-7.5.


emkemp:
Code looks good.  I tested with strict checks and with normal optimization, and behavior looks correct.  I accept.

PR #1628, opened 2024-10-25
Fix the bug in writing model output in 2D ensemble grid space #1627

### LINKED ISSUE #1627: Bug in
### Bug Description

This bug is related to writing model output in 2D ensemble grid space when the map projection is not lat/lon (equidistant cylindrical) and the vertical level (vlevel) is set to 1. In this part of the code, the variable short_name is used instead of dataEntry%short_name. Since the value of short_name is not properly defined, it pulls a random value from memory, which does not match dataEntry%short_name, causing the model to crash.

In the following picture from the TotalView debugger, we can see that short_name is set to "lon," while dataEntry%short_name is "Wind_f." Additionally, dataEntry%varID_def equals 32, which corresponds to "Wind_f."

Screenshot 2024-10-22 at 5 53 41 PM

### Steps to Replicate
II'm not sure if you'll be able to replicate the error because I'm using my branch, which contains all the ICESat-2 project developments.

--assignee @LIS-navari

LIS-navari:
### Description

This bug is related to writing model output in 2D ensemble grid space when the map projection is not lat/lon (equidistant cylindrical) and the vertical level (vlevel) is set to 1. In this part of the code, the variable short_name is used instead of dataEntry%short_name. Since the value of short_name is not properly defined, it pulls a value from memory, which does not match dataEntry%short_name, causing the model to crash.

Resolves #1627

### DIFF
--- a/lis/core/LIS_historyMod.F90
+++ b/lis/core/LIS_historyMod.F90
@@ -66,6 +66,8 @@ module LIS_historyMod
 !                   used in GRIB1 & GRIB2 format
 !  18 Oct 2018: David Mocko: Check lis.config entry for option to turn off
 !                   writing ASCII stats files with netCDF output format
+!  25 Oct 2024: Mahdi Navari: updated the code to fix the bug in writing
+!                    model output in 2D ensemble grid space (see https://github.com/NASA-LIS/LISF/issues/1627 )
 !
 ! !USES: 
   use LIS_coreMod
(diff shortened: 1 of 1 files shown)


emkemp:
The code looks good, and it compiles w/o issue.

I believe @jvgeiger wants to compare this pull request with an earlier one, so I will hold off on merging.

emkemp:
Question raised during today's LIS Tag Up:  Should this change be put into master, or into the public branch first?

LIS-navari:
Hi @emkemp,
I believe this is in the master branch. I don’t think anyone has used this option so far; otherwise, we would have heard complaints about it. It will be included in a future LIS public release.

jvgeiger:
I cherry-picked this commit into the support/lisf-public-7.5 branch.  See pull request #1643.

PR #1630, opened 2024-10-31
Add dynamic lapse rate usage and output to MERRA-2 metforcing

dmocko:
This commit adds the reading and application of pre-generated hourly MERRA-2 lapse rate maps.  This optional config will use these dynamic lapse rates in place of the static lapse rate (of -6.5 K/km) from the "lapse-rate" topographic correction.

Sujay Kumar wrote the initial version of this code, and David Mocko and Sujay worked together to fix some code issues, add code fault tolerance, and add the ability to output dynamic lapse rate on the output grid.

For now, this code should remain in the eis-freshwater2 branch, and should not be merged into the main branch yet.

Credit where credit is due:

@sujayvkumar Added the initial implementation.  See commits b2d501aa5cbd99fc90f114bbaa94a6de662bcba2 and c9ac3fb333b5626931a2ccb96a7a6d0a1b436172

@Kristen-M-Whitney Performed code testing and helped to pre-generate the hourly MERRA-2 dynamic lapse rates.

Testcase to come.

### DIFF
--- a/lis/configs/MODEL_OUTPUT_LIST.TBL.adoc
+++ b/lis/configs/MODEL_OUTPUT_LIST.TBL.adoc
@@ -252,6 +252,9 @@ Density_f:    0  kg/m3   -    1 0 0 1 256 1       # Atmospheric density
 VaporPress_f: 0  -       -    1 0 0 1 256 1       # Vapor pressure
 VaporPressDeficit_f: 0  -  -  1 0 0 1 256 1       # Vapor pressure deficit
 
+#Lapserate
+Lapserate:    1  K/km    UP   1 0 0 1 255 10      # Lapse rate
+
 #Additional FEWSNET Forcings
 PET_f:         0 kg/m2s  -    0 0 0 1 228 1000    # Average PET rate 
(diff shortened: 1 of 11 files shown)


dmocko:
Testcase located here:  /discover/nobackup/projects/lis/tools/pull_requests/1630

jvgeiger:
Hello.  This value, -0.0065, is hard-coded in many places.  Does it make sense to add this to the LIS_constantsMod module?

Kristen-M-Whitney:
Assigning this as a constant makes sense to me. It is the assumed static air temperature lapse rate (-0.0065 K/m), used when we do not utilize the dynamically computed lapse rate.

dmocko:
Agreed.  @jvgeiger  - Would you like to change this, or should I do it?

jvgeiger:
@dmocko, you do it, and I'll approve it. :-)

dmocko:
@jvgeiger 

I have made these changes, and re-ran the testcase.  The output is identical to before.

/discover/nobackup/projects/lis/tools/pull_requests/1630

target    = sub-directory from re-running this testcase
orig_target = sub-directory with original testcase (identical output to "target")
output   = sub-directory where the testcase will write output (only a GrADS .xdf file is within it now).

Issue #1634, opened 2024-11-05
LDT MERRA-2 metforcing unneeded memory allocation and ldtlog issues


dmocko:
The LDT MERRA-2 metforcing reader allocates forcing variables, even if LDT is not going to process metforcing.

This can cause a out-of-memory issues even when in "LSM parameter processing" mode.

Also, some ldtlog messages in this metforcing give partial garbage output.

PR #1636, opened 2024-11-05
Fix LDT MERRA-2 metforcing unnecessary memory allocation and logs

dmocko:
This commit fixes a possible unneeded memory allocation in the LDT MERRA-2 metforcing reader.  Now, the forcing variables are only allocated when LDT is run in one of these three modes:

"Metforce processing"
"Metforce temporal downscaling"
"Statistical downscaling of met forcing"

Also, minor clean-ups are made to log messages for this metforcing.

Resolves: #1634

### DIFF
--- a/ldt/metforcing/merra2/merra2_forcingMod.F90
+++ b/ldt/metforcing/merra2/merra2_forcingMod.F90
@@ -267,19 +267,27 @@ contains
       merra2_struc(n)%startFlag = .true.
       merra2_struc(n)%dayFlag = .true.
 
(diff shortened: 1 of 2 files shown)


jvgeiger:
I agree that (1) could be argued either way.  I will rebase this onto the public branch.

Regarding (2), unfortunately, I do not have an LDT MERRA2 test case.

jvgeiger:
That did not rebase very cleanly.  Because it is not a bug, I will leave it on master.

Issue #1653, opened 2024-12-16
MERRA2 double-sided dynamic lapse rate cutoff


Kristen-M-Whitney:
Option to apply a double-sided cutoff to the MERRA-2 dynamic lapse rate values. 

If using MERRA-2 with dynamic lapse rates turned on, then users can specify the following options in the config file:
- apply a double-sided cutoff - if not present, no cutoff is applied
- specify a maximum lapse rate cutoff value (K/m) - if not present, the default is 0.01 K/m
- specify a minimum lapse rate cutoff value (K/m) - if not present, the default is -0.01 K/m

Note that @dmocko and @sujayvkumar have helped with this enhancement.

emkemp:
The referenced code is not in the master branch. @Kristen-M-Whitney is work still underway on this issue, or can the issue be closed?

dmocko:
This issue is merged into the "eis-freshwater2" branch, and should be closed.

dmocko:
The solution to this issue is merged - not the issue itself.  *long day*

Kristen-M-Whitney:
@dmocko thanks and I agree. Thanks @emkemp for the follow-up!

PR #1661, opened 2024-12-18
Add MERRA-2 dynamic lapse rate double-sided cutoff

Kristen-M-Whitney:
### Description
This commit adds the application of a double-sided cutoff to pre-generated hourly MERRA-2 lapse rate maps. This optional functionality will replace any lapse rate value that falls above (below) the maximum (minimum) cutoff threshold with the cutoff threshold value. The function is invoked by an optional config entry to apply the double-sided cutoff. If this config entry is not present, the cutoffs will not be applied. The commit also includes optional config entries to specify the minimum and maximum cutoff thresholds in units of K/m. If the config entry for the maximum (minimum) threshold is not present, the default is set to +0.01 (-0.01) K/m.

Kristen Whitney and David Mocko worked together to modify the code with this functionality. Kristen Whitney performed code testing.

For now, this code should remain in the eis-freshwater2 branch, and should not be merged into the main branch yet.

Credit where credit is due:
@dmocko helped with the corresponding code modifications and adjustments.

Resolves: #1653

### Testcase
Testcase to come.

### DIFF
--- a/lis/configs/lis.config.adoc
+++ b/lis/configs/lis.config.adoc
@@ -5738,14 +5738,48 @@ Acceptable values are:
 `MERRA2 dynamic lapse rate data directory:` specifies the
 directory of the pre-computed MERRA-2 lapse rate dataset.
 
+`MERRA2 apply double-sided dynamic lapse rate cutoff:` specifies whether
+to apply a double-sided cutoff to the dynamic lapse rate values. This is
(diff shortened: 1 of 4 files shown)


dmocko:
Hi @jvgeiger - since I was involved in this PR, I would like you to review/merge when you have a chance.

Issue #1658, opened 2024-12-17
Openwater model with STATSGO+FAO / Soil texture force exclusion option


slinskik:
### Bug Description

LIS crashes when run with the openwater model over the FLDAS-Global domain (180°W to 180°E, 60°S to 90°N) with STATSGO+FAO soils without setting "Soil texture force exclusion of water points during fill: .true." in LDT when I generate the LIS input file.

Setting "Soil texture force exclusion of water points during fill: .true." fixes the issue.

### Steps to Replicate

@karsenau has replicated this error.

karsenau:
This is issue is addressed with the recent pull request:  https://github.com/NASA-LIS/LISF/pull/1662

With testcase provided to run through the steps.  The case involves running with STATSGOFAO soil texture + 2 surface model types in a simple ensemble case with NoahMP401 + HyMAP, with MERRA2 forcing and perturbed forcing.

PR #1662, opened 2024-12-18
Fix for soil texture force-fill exclusion in LDT

karsenau:
Implemented fix in LDT to address when the
  Soil texture force exclusion of water points during fill is not
  set in the ldt.config file, even though STATSGOFAO soil texture
  and more than 1 surface model type are selected.

 The default value is set to "false", which causes a mismatch for
  points north of -60S in STATSGOFAO soil texture and when surface
  type, e.g., open water, are selected -- and causes LIS run to fail.

 This is issue is documented in the LISF GitHub issue #1658.

 In addition, some additional info has been included in ldt.config.adoc
   and some additional comment notes for SRTM30 native topo readers
   for dealing with points south of -60S.

### Testcase

Test case for this PR and fix can be found here:
/discover/nobackup/karsenau/LISF_PR/LDT_SText_ForceOpt_Fix

Tester is able to run the LDT step in:
/discover/nobackup/karsenau/LISF_PR/LDT_SText_ForceOpt_Fix/LDT_Run

with ldt.config_test

Then run a final test of that file in a LIS test case setup, with the lis.config_test file provided.  All input files are provided.
MERRA2 input forcing is symlinked in directory:  lisdata

### DIFF
--- a/ldt/configs/ldt.config.adoc
+++ b/ldt/configs/ldt.config.adoc
@@ -1633,7 +1633,7 @@ Soil texture spatial transform:     none
 
 `Soil texture fill value for Antarctica:` indicates which soil texture value to be used if an arbitrary value fill is needed south of 60S. If not specified, value of `Soil texture fill value:` will be used.
 
-`Soil texture force exclusion of water points during fill:` indicates whether to force exclusion of water points when filling soil texture. Required when using STATSGOFAO south of 60S (texture is set to all water) and multiple surface model types are used.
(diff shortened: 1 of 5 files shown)


dmocko:
The code compiles, and I was able to successfully run and replicate the ldt and lis testcases.

I approve of the code revisions as well, so will merge this PR.

@karsenau - the lis.config_test file you provided did not have open permissions, but I just used the lis.config_target file, and renamed it to lis.config_test.  I only renamed "target" to "test" within the config file, which made it 6 bytes smaller than the original, which is the same difference as your two lis.config files.  Anyway, just FYI so you can open permissions to your lis.config_test file for others/posterity.  Thanks!

karsenau:
Thanks @dmocko for testing the PR code and use case, and also for mentioning the lis.config_test file permissions.

PR #1672, opened 2025-01-08
Amended FOC files for option to switch to SMAP L2 DA.

emkemp:
### Description

The FOC files are configured to use SMAP_E_OPL.  New SMAP L2 DA settings are added but commented out. Comments are also aded to make changes to the config files to pivot to SMAP L2 if needed.


emkemp:
Good catch.  Corrected.  (I checked the other FOC files and they don't have that error.)

jvgeiger:
Thanks.

PR #1675, opened 2025-01-15
Fix to compile LIS with "WRF coupling: Off"

jvgeiger:
### Description

This fix allows one to compile LIS with `WRF coupling: Off` set in the user.cfg file.

Note that both the GCE coupling runmode and the NUOPC coupling runmode depend on code in the the WRF coupling runmode.  So to use either of those runmodes, you must enable "WRF coupling", which is its default setting.

Note 1) that both the GCE coupling runmode and the NUOPC coupling runmode are not enabled by default and 2) that both require special build commands.

### DIFF
--- a/lis/core/LIS_surfaceModelMod.F90
+++ b/lis/core/LIS_surfaceModelMod.F90
@@ -8,6 +8,7 @@
 ! All Rights Reserved.
 !-------------------------END NOTICE -- DO NOT EDIT-----------------------
 #include "LIS_misc.h"
+#include "LIS_plugins.h"
 ! Macros for tracing - Requires ESMF 7_1_0+
 #ifdef ESMF_TRACE
 #define TRACE_ENTER(region) call ESMF_TraceRegionEnter(region)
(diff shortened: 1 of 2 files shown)


emkemp:
LIS compiles w/o issue.  Code looks fine.  I accept.

dmocko:
I am trying to compile the code with the "-2" configure option on the Milan nodes using lisf_7.5_intel_2023.2.1_mapl (using "strict-checks").  However, I am receiving this error:

```
icx: error: unknown argument '-check=conversions,stack,uninit'; did you mean '-fcheck=conversions,stack,uninit'?
icx: error: unknown argument '-ftrapuv'; did you mean '-ftrapv'?
icx: remark: Note that use of '-g' without any optimization-level option will turn off most compiler optimizations similar to use of '-O0'; use '-Rno-debug-disables-optimization' to disable this remark [-Rdebug-disables-optimization]
```

My make/user.cfg is:

```
Flake.1.0: Off
VIC.4.1.1: Off
VIC.4.1.2: Off
CLM.2: Off
LSM RDHM.3.5.6: Off
AWRAL: Off
AWRAL.6.0.0: Off
```

The above error happens after trying to compile: `runmodes/wrf_cpl_mode/LISWRFimport_module.F90`

If I add this to my make/user.cfg as well:

`WRF coupling: Off`

..then the error happens when trying to compile: `core/LIS_RTM_FTable.c`

Any advice on what's happening?

PR #1693, opened 2025-03-18
Fixed bug in reporting check for prior 12z analysis.

emkemp:
### Description

Fixed bug in reporting check for prior 12z analysis.

This will change the final result of USAFSI.  With the bug, at 00Z, 06Z, and 18Z USAFSI will erroneously search for (nonexistant) 0.25 deg SNODEP data, and then apply climo data where |lat| > 40 deg to fill data voids or where zero snow is analyzed.  No change in behavior at 12Z, since USAFSI will actually check for (and should find) the prior 12Z analysis for updating snow and ice age.


emkemp:
This patch needs to be migrated to LISF 7.6 and master after being applied to LISF 7.5.

PR #1716, opened 2025-07-08
SNIP runmode for LDT

emkemp:
### Description

Created new SNIP runmode for LDT, cannibalizing and renaming code from USAFSI.

This is intended to allow code changes for SNIP specific algorithms and internal workflow without breaking the USAFSI runmode.

This is not finished yet.


emkemp:
I can reproduce Kehan's test case.  Since @jvgeiger is not available, I will accept my own pull request.

PR #1721, opened 2025-07-18
SNIP DA plugins for Noah39 and NoahMP401

emkemp:
### Description

SNIP DA plugins for Noah39 and NoahMP401


emkemp:
Note that this pull request is *not* ready yet.  It will not be complete until the LDT/SNIP output is finalized, but we at least have a running start.

emkemp:
I completed test runs from 1-11 Jan 2025 for both Noah39 and NoahMP401.  Each LSM had two runs, one assimilating USAFSI, the other assimilating SNIP proxies which were just symbolic links to the USAFSI files.  The output files were identical for the same LSM.  Since the logic of the SNIP plugins is identical to the USAFSI plugins (except for the file names), I believe the code can be trusted.  

I am therefore removing the DRAFT designation for this pull request.

Issue #398, opened 2019-11-04
LIS Irrigation output log messages needed


karsenau:
Currently, LIS can only generate daily irrigation output (i.e., not monthly) and only runs if irrigation is written out when irrigation is turned on (selected on in the MODEL_OUTPUT table).  

No alerts are provided to the user who may set such options. The following two situations require alerts in LIS to the users, who need to run with irrigation turned on:

1) When irrigation is turned on in the LIS config file but not activated in the MODEL OUTPUT table, LIS simply crashes with no message given. Conditional statements need to be added to let user know.

2) Currently, LIS will run with the "Irrigation output interval:" set to "1mo" (monthly interval), but two issues are encountered:  all 0's are written out and the month-date in the file naming convention is generated incorrectly. Also here, conditional statements are needed to ensure the user does not select this option, until a fix is implemented.

Finally, LIS irrigation documentation needs a couple of minor updates and additional details.

karsenau:
Check to see if this was ever fixed ...

hkato1:
Hi @karsenau 

Thanks for bringing this up.  I had tried daily and sub-daily irrigation output intervals before and never monthly.  In the pull request I just submitted, I was able to address them.

1.  If irrigation is turned on in the LIS config but not activated in the MODEL OUTPUT, the run crashes with the message in log file.  [ERR] 'Irrigated water:' was not selected. Program stopping ...
[ERR] endrun is being called

2. Monthly interval seemed to be working, maybe it got updated somewhere along since 2019 :)  However, it was using the generic "30-day" window.  So the monthly files are output on Jan 2, Jan 30, Mar 2, Apr 1, May 2... and so on.  I fixed it so it creates monthly output at the last model time step of the month (e.g. Jan 31 at 23:45 for 15-min time step).  This way, you get one output per month.  Note testing configuration needs to capture the domain that's irrigated and long enough period to activate irrigation.

Please let me know if you have any problems or suggestions.

emkemp:
This issue is still valid, awaiting resolution of PR #1762.

Issue #177, opened 2019-06-25
Add Irrigation Routines to Noah 3.9


jerlingi:
Add Noah 3.3 irrigation interfaces to Noah 3.9

emkemp:
This issue is still active. Awaiting resolution of PR 1762.

Issue #176, opened 2019-06-25
Add Irrigation Routines to Noah 3.6


jerlingi:
Add Noah 3.3 irrigation routine interfaces to Noah 3.6

dmocko:
Hi @jerlingi 

We have a user trying to do this right now in Discussion #1584 

Is there more to adding the irrigation that I may have missed in my answers on this discussion?

I do note that this issue probably refers to all types of irrigation, unlike the discussion, which is only about sprinkler.

jerlingi:
Hi @dmocko 

I reviewed your discussion on #1584 and my pull request from when I put irrigation into Noah-MP 4.0.1, and you have identified the three files that we need to see to help this user debug:

LISF/lis/make/default.cfg
LISF/lis/plugins/LIS_lsmirrigation_pluginMod.F90
LISF/lis/surfacemodels/land/noah.3.6/irrigation/noah36_getirrigationstates.F90

I'll keep an eye on this discussion. Copying the Noah 3.3 routines over, renaming them, and making sure default.cfg and LIS_lsmirrigation_pluginMod.F90 have been correctly updated should allow all irrigation types to be enabled in Noah 3.6.

emkemp:
This issue is still active, awaiting resolution of PR 1762.

Issue #107, opened 2019-04-25
Add multiple irrigation scheme support in LIS


smahanam:
This is part of the same task "improving the LIS irrigation scheme to support multiple crop types and multiple irrigation types" that Hiriko, I and Kristi are working on.

We have created new parameters that are needed for the proposed irrigation model improvements on the NLDAS domain (and Global as well). While upgrading LDT is underway, we want to start working on the LIS side by reading the new parameter file into LDT “as-is”, and process it with Noah-3.3 LSM parameters for the NLDAS domain via LDT and start testing with with Noah3.3

Thank you !
Sarith

smahanam:
NOTE that this is just the 1st attempt not for merging.
I have made 2 changes:
1) Added "concurrent" irrigation (per Sujay's suggestion) - a new 4th irrigation type. We compute weighted average of Sprinkler, Flood, and Drip
2) Added Drip irrigation (based on what Ben had done years ago)

I have committed to the feature/irrigation branch on my fork
smahanam/LISF-1

https://github.com/hkato1/LISF/compare/feature/irrigation...smahanam:feature/irrigation?expand=1

dmocko:
Hi @hkato1, @smahanam, and @karsenau

Is there any update on this?

Can this issue be closed, if it is the same as https://github.com/NASA-LIS/LISF/issues/454 and https://github.com/NASA-LIS/LISF/pull/455?

emkemp:
This issue is still active, awaiting resolution of PR 1762.

PR #1762, opened 2026-01-09
Feature/irrigation

hkato1:
### Description

Enhancements are made for the irrigation modeling in LIS and preprocessing irrigation related datasets in LDT.  We added multiple crop and multiple irrigation type support in LDT.  We updated each irrigation type currently enabled in LIS and added multiple irrigation scheme where Sprinkler, Drip, Flood, and Paddy irrigation types occur simultaneously.  We updated irrigation code structure and consolidated routines.  This implementation is based on Rodell et al. 2026 submitted to JAMES.  Currently works with Noah LSM and will be incorporated into NoahMP and others next.

Resolves original issues #106 and #107 at last!  Also resolves related issues #176, #177, and #398.

Contributed by @smahanam, @karsenau, @jerlingi, and @KathyNie.  Thank you!

### Testcase

/discover/nobackup/hkato/IRRIGATION/testcase/issue106
/discover/nobackup/hkato/IRRIGATION/testcase/issue107


hkato1:
> Hi @hkato1
> 
> I compiled and ran the LDT and LIS executables to compare with your sample output. In the LDT output I see some isolated pixels with noise-level differences, which I don't think is a problem. Similar story for the LIS IRRIGATION output.
> 
> For the LIS SURFACEMODEL output, I also see noise-level differences but over wide areas. Enough so that I'm concerned your output was generated using different compiler settings. What compiler level did you use? Is this 2 or 0 or ???

Hi @emkemp

Thanks for working on the PR.  I compiled with -1 (debugging) level.

hkato1:
> @hkato1 It has been a while, so I forget. I believe the new irrigation_model offers a configuration to mimic simple physics/modeling approach of Mutlu's irrigation model. However, that will not produce identical results of Mutlu's model. Am I right?

@smahanam yes, we can configure different irrigation approaches in the new irrigation model.  Mutlu's approach was using Sprinkler only, soil deficit demand, and greenness fraction based growing season determination--which can be set in lis.config.  The result should be close.  Since we are now using irrigation class (i.e. IM) and moved many variables into lis.config, the interface and main LSM routines need to be updated to use that new structure.

jvgeiger:
I must go through your history to see why the FOC files were deleted.

Did you purposely make changes to read_gdas.F90 and read_princeton.F90?

PR #1766, opened 2026-01-16
Correct writing 1-D lat/lon values into routing output file

jvgeiger:
### Description

Fix the writeroutingvar_netcdf_real routine to build the 1-D lat/lon arrays that are written into the routing output files, which is enabled when setting

"Number of dimensions in the lat/lon output fields:" to "1D".


emkemp:
Code compiles w/o issue.  Intel Fortran compiler produces warnings with strict checks turned on; however, these warnings are not in the new code segments.  I accept.

PR #1785, opened 2026-03-05
Changes to support PotEvap_tavg from NoahMP401.

emkemp:
### Description

This includes a patch to NoahMP401_main.F90 to check for missing values.


emkemp:
I believe this is ready for review.

dmocko:
I will review - and share more detailed evaluations.  Based on your "ready for review", I am interpreting that we are going to recommend turning on PotEvap output in noahmp401 in LIS via the MODEL_OUTPUT_TBL - because LVT will be expecting this variable for post-processing.  I agree with this sentiment, as the variable output is of sufficient quality.

emkemp:
Hi @dmocko 

Yes, @jerrywegiel gave the go-ahead via e-mail.  Full support all the way through postprocessing, so that Noah and NoahMP output the same variables.

dmocko:
Evaluation/comparisons of both this Noah-MP-4.0.1 PotEvap and Noah-3.9 PotEvap against the satellite-observation-based GLEMA PotEvap product shows that the magnitudes/patterns from Noah-MP PotEvap are reasonable.

I will accept this PR.

Note that when the final FOC files and MODEL OUTPUT TBL files are delivered, the "PotEvap" variable will need to be turned on within the Noah-MP-4.0.1 LIS output.

PR #1786, opened 2026-03-09
Fix divide-by-zero

jvgeiger:
### Description

Fix a divide-by-zero error raised by the GNU compilers.  This error was not raised by the Intel nor Cray compilers.


emkemp:
@jvgeiger 

The code compiles w/o issue.

I note, however, that the indentation of this file is messed up, and Intel's strict checks complains about a few things.  Since these issues are beyond the scope of a bug fix, I will accept as is.

PR #1750, opened 2025-11-25
Add support for SWOT–HyMAP DA

yyoon4:
### Description

Add support for SWOT–HyMAP2 DA (#1749)

Currently, the SWOT DA code supports only HyMAP2. Additional code for HyMAP3 will be submitted later.

### Testcase
/discover/nobackup/yyoon4/lis7/Testcase/swot_da

Issue #1749, opened 2025-11-25
HyMAP SWOT DA


yyoon4:
**Develop the assimilation of SWOT Water surface elevation (WSE) into the Hydrological Modeling and Analysis Platform**

SWOT observation and assimilation workflow
<img width="952" height="294" alt="Image" src="https://github.com/user-attachments/assets/e471475b-85c8-4d55-8db6-2e7b932da563" />

The DA code includes a localization scheme

Please refer to the detailed algorithms and results in this draft:
https://essopenarchive.org/doi/full/10.22541/essoar.176361210.02941601/v1

emkemp:
This is in a holding pattern, waiting for PR 1798 to be resolved (I think we're basically there, but @jvgeiger and I need to review it one more time).

PR #1798, opened 2026-04-13
Support SWOT-DA for HYMAP3

yyoon4:
### Description

SWOT-based data-assimilation (DA) support for the **HYMAP3** routing module (#1749, #1750) 

### Testcase

/discover/nobackup/yyoon4/lis7/Testcase/swot_da_hymap3

### DIFF
--- a/lis/configs/lis.config.adoc
+++ b/lis/configs/lis.config.adoc
@@ -11174,6 +11174,20 @@ Acceptable values are:
 |1     | Use 
 |====
 
+`HYMAP3 use localization update in DA:` specifies whether to enable the localization scheme for DA.
+
+Acceptable values are:
+
+|====
+|Value   | Description
+|0       | Do not use localization
+|1       | Use localization for DA
+|====
+
+`HYMAP3 localization update window size:` specifies localization update window size.
+
+`HYMAP3 localization weight map:` specifies the file location of the localization weight map
+
 .Example _lis.config_ entry
 ....
 HYMAP3 routing model time step: "30mn"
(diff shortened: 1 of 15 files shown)


emkemp:
Hi @yyoon4 

I reproduced the results from your test case.  Hooray!

Question:  Is this PR intended for the LISF 7.8 557WW release, or is this for another project?

yyoon4:
No, 

> Hi @yyoon4
> 
> I reproduced the results from your test case. Hooray!
> 
> Question: Is this PR intended for the LISF 7.8 557WW release, or is this for another project?

No, it is not intended for the LISF 7.8 557WW release. It is just for the master branch and for use in the SWOT project.

emkemp:
Hi @yyoon4 

At this point, we have two remaining issues with this pull request.

(1) The lis.config entries for da_SWOT in HYMAP3 are identical to the entries for da_ML.  Does this need to change, i.e., should the config options mention "SWOT" or "ML" as well as "HYMAP3" for clarity?

(2) Can you review the existing LIS user guide and update as necessary for this PR?

Thanks.

yyoon4:
@emkemp 

Thanks for checking and cleaning code.

da_WL is intended for synthetic water level DA. I believe this code was copied from HyMAP2, but it has not been tested, so I am not sure whether it works correctly. The da_WL code is generally similar to da_SWOT; however, da_SWOT has since been updated to support SWOT observations and the new localization approach. For now, I believe it would be better to remove the directory from HyMAP3.

For lis.config, the basic concept is the same as for SWOT, so the configuration entries are similar. 

However, da_SWOT uses:
Data assimilation set:                    **"SWOT water level"**

I revised the lis.config file as follows:

Observation perturbation algorithm:       "GMAO scheme"
Observation perturbation frequency:       "1hr"
Observation attributes file:              ./input/da/**swotobs_attribs.txt**
Observation perturbation attributes file: ./input/da/**swotobs_pertattribs.txt**

I also updated 'lis.config.adoc' file

emkemp:
Thanks @yyoon4.

@jvgeiger:  Any objections to me removing lis/routing/HYMAP3_router/da_WL?

emkemp:
@yyoon4 We are dusting off this PR.  It appears the HyMAP3 da_WL component was committed by @agetiran, so I am now opposed to removing it.

The three common lis.config entries are

HYMAP3 use localization update in DA:
HYMAP3 localization update window size:
HYMAP3 localization weight map:

And this same pattern of shared lis.config entries is also present in HyMAP2 for da_WL and da_SWOT.  So I'm inclined now to just accept it.

PR #1806, opened 2026-04-29
FOC files for LISF 7.8

emkemp:
### Description

FOC files for LISF 7.8.


dmocko:
Hi @emkemp 

Thanks for adding this draft PR.  I will take a closer look at the NRT FOC files tomorrow.

I recommend removing the "README", "URBPARM" and "_UMD.TBL" files, which aren't needed.

Also, it may be better to move the "noah39_parms" and "noahmp401_parms" directories for each sub-system
into a single "LS_PARAMETERS" directory on the same level as MR, NRT, etc.  That way there's less confusion,
and only one set of these files.

dmocko:
Should we rename the two ldt.config files from "v761" to "v780" in this directory?

Or just "v78" to be consistent with other configs in the "557WW-7.8-FOC" sub-directory?

lis/configs/557WW-7.8-FOC/7.8.0/NRT/ldt_lis_param

dmocko:
"snip_attribs.txt" and "snip_pertattribs.txt" could be edited to change "USAFSI" to "SNIP".

However, it's just for documentation and this change is not required to run the snow DA with SNIP.

dmocko:
Line 77 (commented out) could be removed from the three lis.config.nrt_lis_noah39* files.

Line 124 has a typo - "netdf" instead of "netcdf" - in the two lis.config.nrt_lis_noah39*snip* files.

dmocko:
Other than above, the lis.config.nrt_lis* files all look good, as well as the files in "attribs", "ldt_lis_param", "ldt_misc", "tables", and the files in the "_parms" directories.

I did not check files in "smap_e_opl" or "wsf_m_opl", and will leave those for others to confirm.

emkemp:
I think we're ready for this PR.

PR #1810, opened 2026-06-05
Create NLDAS-3 forcing directory structure

fadjimaina:
This PR creates initial NLDAS-3 forcing directory for the retrospective.
It includes subfolders for each forcing variables


dmocko:
Hi @fadjimaina 

Thanks for starting the framework for everyone's code contributions.

A minor suggestion before accepting this PR:

Remove the "humidity" and "temperature" sub-directories, and add a "surface" sub-directory.
The "surface" will cover the temperature, humidity, and surface pressure.  These variables are
all processed simultaneously, so I think they can go together here.

@Kristen-M-Whitney - does this seem reasonable to you?

Kristen-M-Whitney:
@dmocko Yes, that makes sense to me.

One caveat is that I think the only *new* code that would belong in `lis/utils/nldas-3_forcing/retrospective/surface` is the code used to compute the dynamic lapse rates.

The actual surface forcings were generated using functionality that is already part of the `eis-freshwater2` branch of LISF. Specifically, LIS was used to downscale MERRA-2 surface temperature, humidity, and pressure to 1 km following the Whitney et al. (2025) approach (i.e., the Cosgrove method with dynamic lapse rates).

In addition to the dynamic lapse-rate code, I also have:

* LIS configuration files used for the downscaling runs
* LIS job scripts used to execute the runs
* Scripts used to merge the hourly LIS outputs into a single daily file containing all 24 hours

Would we want to include those configuration/job files and the daily-merging scripts in this directory as well?

I also suggest documenting in the README for both this directory and `lis/utils/nldas-3_forcing/retrospective/LWdown` that the downscaled MERRA-2 surface fields (generated through the Whitney et al. (2025) workflow) were subsequently used in the LWdown bias-correction process for the interpolated CERES data.

@dmocko and @fadjimaina, does that sound reasonable?

For the `surface` subdirectory, I could see including:

* Dynamic lapse-rate generation code
* LIS configuration and job files used for the MERRA-2 downscaling workflow (Whitney et al., 2025; `eis-freshwater2` branch)
* Scripts used to merge hourly LIS outputs into daily files

Thoughts?

dmocko:
Hi @Kristen-M-Whitney 

Yes - this sounds good to me.  Definitely yes to including the dynamic lapse-rate generation code,
the lis.configs, and the merging scripts.  Up to you if the job scripts are needed.  If they are very
basic (aka, they just submit LIS), then probably not needed.  But if you have other lines to handle
the data/setup/etc., they please include them.  The README should be updated to explain the full
process.  

For clarification, the path you mention above should be: lis/utils/nldas-3_forcing/retrospective/lwdown

I like all lower-case with directory names when possible (easier to type).

Kristen-M-Whitney:
@dmocko ok that all makes sense, thanks!

fadjimaina:
I've updated it

fadjimaina:
Yes, I think we don't need scripts to run the models, only lis.config files if those are necessary

dmocko:
Thank you @fadjimaina 

Looks great.  I will merge.

PR #1811, opened 2026-06-05
Add NLDAS-3 retrospective precipitation generation codes

fadjimaina:
Added fortran codes required to generate NLDAS-3 retrospective precipitation.


emkemp:
Hi @fadjimaina 

How do you compile these programs?  Neither './compile' nor './compile -u' work, and there is no Makefile in the directory.

fadjimaina:
@emkemp I am adding the Makefile and compilation instructions.

emkemp:
Hi @fadjimaina:

I worked with your codes to remove unused variables, improve the looping to avoid cache misses, and revised the indentation.  (I did not address warnings about MPI routines because that a known problem with Intel MPI.)

Can you rerun these codes to verify the answers are not affected?  Or alternatively, can you provide test cases so I can verify them myself?

PR #1819, opened 2026-06-17
Fix incompatible function pointer types gcc 15+

### LINKED ISSUE #1818: Incompatible function pointer error gcc v15+
### Bug Description

gnu23 standard used in gcc 15+ will no longer treat  `void (*func)();` as an unspecified number of arguments. It's now identical to `void (*func)(void);` but the call to this function `call optuereadrestart(trim(LIS_rc%optuealg)//char(0), rstflag)` includes a rstflag argument. The `struct optuerstnode` needs to specify arguments in the func variable.

### Steps to Replicate

Compile with gcc 15

danrosen25:
### Description

Resolves #1818 

- Update the restart registry callback type from void (*func)() to void (*func)(void*) so it matches the registered function signature.
- Change FTN(optuereadrestart) to accept a restart flag pointer (void *rstflag) and forward it to the callback.
- Replace the callback invocation from current->func(); to current->func(rstflag);.

### Testcase
Needs proper testing. Compiles fine with gcc-15

### DIFF
--- a/lis/core/LIS_optUEMod.F90
+++ b/lis/core/LIS_optUEMod.F90
@@ -275,10 +275,8 @@ contains
 !  \end{description}
 !EOP    
     implicit none 
-    
-    logical         :: rstflag
 
-    call optuereadrestart(trim(LIS_rc%optuealg)//char(0), rstflag)
+    call optuereadrestart(trim(LIS_rc%optuealg)//char(0))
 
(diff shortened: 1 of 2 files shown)


jvgeiger:
Thanks @danrosen25 for finding this error.  The callback routines do not use the rstflag argument.  I corrected the generic call to optuereadrestart, and I removed the rstflag from the C code.

PR #1820, opened 2026-06-18
Add pio option to configure

### LINKED ISSUE #1744: Turning on PIO Support within ESMF
In order to support the default build for ESMF and support PIO dependent ESMF features we can add -lpioc to the linker flags using the below changes to `Config.pl`. A more robust solution would include scanning the `esmf.mk` to see if PIO was enabled within ESMF. There's an example of reading the `esmf.mk` for `use_esmf_trace`.

Here's the simplest solution, which works because the default build for ESMF installs lpioc into the LIB_ESMF directory.

```perl
print "Use PIO? (1-yes, 0-no, default=0): ";
$use_pio=<stdin>;
chomp($use_pio);
if($use_pio eq ""){
   $use_pio = 0;
}
```

or we can use an environment variable that doesn't prompt the user
```perl
# USE_PIO does not prompt user
if ($ENV{USE_PIO} eq '1') {
   $use_pio = 1;
}
```

```perl
if($use_pio == 1){
   $ldflags = $ldflags." -lpioc";
}
```

danrosen25:
### Description
Resolves #1744

Add `Use PIO?` option to Config.pl. The default for this option is `0-no`. PIO location can be defined using `LIS_PIO` or setting `LIS_PIO_IN_ESMF = 1`. The latter will use the ESMF library location to set the PIO library and include locations. If neither environment variable is set than Config.pl throws an error.

### Testcase
Build LIS on Discover including the esmf/8.6.1_intel-2023.2.1_impi-2021.11 library
`/discover/nobackup/projects/lis/libs/sles-15.4/esmf/8.6.1_intel-2023.2.1_impi-2021.11/`

### DIFF
--- a/docs/LIS_users_guide/build.adoc
+++ b/docs/LIS_users_guide/build.adoc
@@ -48,6 +48,7 @@ Set the appropriate LIS environment variables.  These are used to control the co
 | `LIS_PETSC`     | path to PETSc library            | optional
 | `LIS_JPEG`      | path to JPEG library             | optional (use system libjpeg by default)
 | `LIS_RPC`       | path to RPC library              | optional (use system RPC library by default)
+| `LIS_PIO`       | path to PIO library              | optional (only for NASA Land Coupler (NLC) and NASA Water Insight projects)
 |====
 
 Set the `LIS_ARCH` environment variable based on the system you are using.  The following commands are written using Bash shell syntax.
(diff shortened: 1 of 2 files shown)


danrosen25:
@jvgeiger Thanks for adding documentation. I was thinking about that today.

PR #1860, opened 2026-09-08
Clean up example file following guidance

jvgeiger:
### Description

Clean up example file following guidance


smahanam:
Thank you @jvgeiger !

PR #1666, opened 2024-12-28
LIS/Noah-MPv5.0 coupling

cenlinhe:
This PR is to submit the code update of coupling Noah-MPv5.0 with LIS through the GitHub submodule mechanism.

- The refactored Noah-MP source code under LISF/lis/surfacemodels/land/noahmp.5.0/phys/noahmp is currenlty directly linked to the official community version of NCAR's Noah-MP GitHub: https://github.com/NCAR/noahmp, branch: release-v5.0-LIS
- The main change of this PR is to create a new noahmp.5.0 folder under LISF/lis/surfacemodels/land/ to couple with the Noah-MPv5.0, with some changes in the lis driver level and documentation part to work with the Noah-MP v5.0.
- The model code has been tested, benchmarked, and evaluated by the NCAR team in the NCAR's HPC (Derecho) and has recently been tested in NASA's HPC by David Mocko for both LSM run and DA of soil moisture and snow.
- The new lis.config files for Noah-MPv5.0 are also provided as examples under LISF/lis/configs/Noah-MPv5.0/
- The model benchmark and evaluation results are currently submitted to _Geoscientific Model Development (GMD)_ through the manuscript of "Benchmarking and evaluating the NASA Land Information System coupled with the refactored Noah-MP land surface model version 5.0"
- The new Noah-MPv5.0 physics options are described in the LISF/lis/configs/lis.config.adoc
- Due to the GitHub submodule mechanism, users need to use "git clone **--recurse-submodules** https://github.com/NASA-LIS/LISF" to download both LIS and Noah-MP source code. This needs to be added to the LIS Readme file.

Code contributor: Cenlin He (NCAR; cenlinhe@ucar.edu)
Model test contributors: Cenlin He (NCAR), Tzu-Shun Lin (NCAR), David Mocko (NASA/LIS)

PR #1862, opened 2026-09-09
Add Noah-MP-5.0 LSM parameter processing to LDT

dmocko:
This pull request adds LDT parameter processing for the Noah-MP-5.0 LSM.

See also the pull request for adding the Noah-MP-5.0 LSM into LIS - #1666.

### DIFF
--- a/ldt/configs/ldt.config.adoc
+++ b/ldt/configs/ldt.config.adoc
@@ -89,6 +89,7 @@ Surface model types:  "LSM"
 |Noah.3.9 |Noah 3.9
 |Noah-MP.3.6 |Noah-MP 3.6
 |Noah-MP.4.0.1 |Noah-MP 4.0.1
+|Noah-MP.5.0 |Noah-MP 5.0
 |CLM.2 |CLM version 2.0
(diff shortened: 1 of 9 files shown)


dmocko:
Testcase for this pull request is found here:

/discover/nobackup/projects/lis/tools/pull_requests/1862

dmocko:
@emkemp - I am fine with these changes.  However, they are not to the sections of the code that I made for this PR.  In other words, the problems that you have fixed already existed in these files before this PR.

Are we cleaning up codes as we go along, or should these types of code cleanups be in separate PRs going forward?

emkemp:
@dmocko I agree the changes do not affect the code updates that you made for the PR.

My philosophy is "we need to clean all these codes up"; and if a particular file is part of a PR and needs to be examined,  it's worth cleaning up.  Kill two birds with one stone.

I note there are no conflicts reported with master, so there seems to be no reason to revert my changes.

dmocko:
@emkemp - agreed.  I am fine with cleaning up these codes as we go when they are part of a PR.  Thanks for the response.

